PR #15, opened 2019-08-12
Add keyvault and labapi to settings.gradle

shahzaibj:
In this PR, I've added the keyvault and labapi projects to the global settings.gradle file so that they can be used in other subprojects as well. This allows us to utilize the keyvault and labapi for tests that are written in common, or msal or any other submodule. 

keyvault and labapi have also been moved from adal to common. You can take a look at the following PRs.

adal - https://github.com/AzureAD/azure-activedirectory-library-for-android/pull/1456
common - https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/609

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -10,9 +10,7 @@
 rootProject.name = 'android_auth'
 
 include ':adal', ':msal', ':common', ':AADAuthenticator', ':brokerHost', ':msalTestApp', ':adalTestApp',
-        ':app', ':library', ':NgcProviderLibrary', ':AadRemoteNgcLibrary', ':SharedCoreLibrary',
-        ':keyvault', ':labapi'
-        ':app', ':library', ':NgcProviderLibrary', ':AadRemoteNgcLibrary', ':SharedCoreLibrary', ':AzureSample', ':TestUtilitiesLibrary'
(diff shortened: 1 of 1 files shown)


settings.gradle
```diff
@@ -23,5 +23,7 @@ project(':library').projectDir = new File ('authenticator/PhoneFactor/OpenSource
 project(':NgcProviderLibrary').projectDir = new File ('authenticator/PhoneFactor/NgcProviderLibrary')
 project(':AadRemoteNgcLibrary').projectDir = new File ('authenticator/PhoneFactor/AadRemoteNgcLibrary')
 project(':SharedCoreLibrary').projectDir = new File ('authenticator/PhoneFactor/SharedCoreLibrary')
+project(':keyvault').projectDir = new File('adal/keyvault')
```
rpdome: shouldn't this be under common?

shahzaibj: I've added them to the global settings.gradle file so that they can be imported into other projects. I'm already importing them into common and msal as follows

```
    implementation project(':keyvault')
    implementation project(':labapi')
```

shahzaibj: As discussed, I will be removing the keyvault and labapi from adal and move them to common. Once that is done I will create PR for those and link those here as well.

PR #35, opened 2020-05-18
Add more details

rpdome:
Got a feedback that the folder could be hidden.


rpdome:
> Bit of a nitpick for such a small diff, but adding more details to what?
 
details added. thanks.

PR #36, opened 2020-06-03
Remove unused authenticator values

rpdome:
These values are copied from authenticator a while back. 

Now, android_complete is configured in such a way that it would just read these values from authenticator's version file during build time, so these values are no longer needed.

(NOTE: the values here are outdated and it would crash authenticator app)

### DIFF
--- a/gradle/versions.gradle
+++ b/gradle/versions.gradle
@@ -11,15 +11,6 @@ ext {
     gradleVersion = '3.6.1'
     androidMavenGradlePluginVersion = "1.4.1"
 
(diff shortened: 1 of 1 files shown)


shahzaibj:
I actually haven't noticed Auth app crash for me lately

rpdome:
> I actually haven't noticed Auth app crash for me lately

Probably depends on what commit you are on. They recently shipped a bug and fix it by reverting back Kotlin version. Android_complete still have the old (higher) version, hence it crashed on my machine.

PR #68, opened 2021-03-19
Add doc on gradle project properties and cmd line args

shahzaibj:


AdamBJohnsonx (approved):
Could we put a link to this file from the root README.md?

tanmaymanolkar1 (approved):

shahzaibj:
> Could we put a link to this file from the root README.md?

Addressed here: https://github.com/AzureAD/android-complete/pull/68/commits/c4040d4dd69e5a41922fb90f247d4873a844a004

docs/ProjectBuild/gradle_project_properties.md
```diff
@@ -0,0 +1,50 @@
...
+
+![Android Studio Command Line Parameters](command_line_args.png "Android Studio Command Line Parameters")
+
+## Which properties / arguments are supported across which of our android-complete modules?
+
+We currently support the following gradle properties via command line across the following modules:
+
+| Property               | Declaring Module                              | Description                                                  |
+| ---------------------- | --------------------------------------------- | ------------------------------------------------------------ |
+| labSecret              | testutils                                     | Enables access to [Lab Api](../Automation/labsetup.md) to get test accounts |
```
iamgusain, line 31: it will be good to add another column that shows example/supported values for these properties

shahzaibj: I've done it for brokerSource here: https://github.com/AzureAD/android-complete/pull/68/commits/dd4827c0f6fab443353f9e3111e6d9255cae1c34

For the rest of these, the supported values are self explanatory such as msalVersion has to be a valid msal version...lab secret needs to a client secret value etc...

I do have an example of how to pass these at the bottom of this doc....for all these the mechanism to supply them is the same so I think one example should be sufficient.

PR #71, opened 2021-04-07
Adding plugin

shoatman:
I added a plugins folder to hold 1 or more plugin projects.  I also added documentation for the plugin to the plugin project.


plugins/infra/src/main/java/com/microsoft/identity/infra/BuildPlugin.java
```diff
@@ -0,0 +1,39 @@
...
+
+        final BuildPluginExtension config = project.getExtensions()
+                .create("aca", BuildPluginExtension.class);
+
+        project.afterEvaluate(project1 -> {
+            if(config.getDesugar().get()) {
+                project1.getLogger().warn("DESUGARING ENABLED");
+                applyDesugaring(project1);
+            }else{
+                project1.getLogger().warn("DESUGARING DISABLED");
```
AdamBJohnsonx, line 23: I'd think we'd want to affirmatively set at least the coreLibraryDesugaringEnabled to false here.

shoatman: Strictly speaking that would be the default.... I think your concern raises perhaps a more important question regarding the required version of the android gradle plugin in order for this work at all.  I'll think on that as well.

shoatman: I'm going to fix this later.

PR #76, opened 2021-04-23
Add common-java

rpdome:
related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1300

Given that common-java are being consumed as source, android complete will not (yet) break *without* this change.

At some point, once we switch to the 'dependency' model, then this will be required.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -19,13 +19,14 @@ pluginManagement {
 rootProject.name = 'android_auth'
 
 // Android DevX Projects
-include ':adal', ':msal', ':common', ':AADAuthenticator',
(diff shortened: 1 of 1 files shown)


gitaumoses4 (approved):

rpdome:
Local MSAL is consuming dist common - and it's breaking... due to a breaking change in logger (MSAL is now consuming `com.microsoft.identity.common.logging.Logger`, not `com.microsoft.identity.common.internal.logging.Logger`)

We might want to use either use local on both for verification.

cc: @iambmelt, @p3dr0rv

PR #78, opened 2021-04-27
Allow plugin to set JAVA 8 for JAVA only projects

shahzaibj:
Also add plugin to android complete project

Related PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1338

Also deleted the `infra` directory. Had chat with @shoatman during scrum and we determined that `infra` directory is not needed..we only need `buildsystem`


plugins/buildsystem/src/main/java/com/microsoft/identity/infra/BuildPlugin.java
```diff
@@ -36,4 +41,11 @@ private void applyDesugaring(final Project project){
         });
 
     }
+
+    private void applyJava8ToJavaProject(final Project project) {
+        project.getPluginManager().withPlugin(JAVA_LIBRARY_PLUGIN_ID, appliedPlugin -> {
+            project.setProperty(JAVA_SOURCE_COMPATIBILITY_PROPERTY, JavaVersion.VERSION_1_8);
```
rpdome, line 47: Why is this differently set from the function below?

shahzaibj: Because this is a java project as opposed to Android projects...Android projects can be either an app or a library and Android has built these extensions using the Android plugin and moved the setting of these properties over to that extension...so for Android we check specifically for that extension and set property to that....however the JAVA only project doesn't have that concept and source compatibility is just a gradle project property directly on the project itself...so we just set it in this way.

shoatman (approved):
lgtm.  Thanks Shahzaib.

PR #79, opened 2021-04-28
Update azure-pipelines.yml 

p3dr0rv:


shahzaibj:
Not to clear to me why the resource reference was unnecessary...can you elaborate?

p3dr0rv:
> Not to clear to me why the resource reference was unnecessary...can you elaborate?
@shahzaibj 
 even If you don't explicitly declare the resource reference for the own repo in which the  pipeline is located,  your source code is checked out into a directory

azure-pipelines.yml
```diff
@@ -4,14 +4,8 @@
 
 name: Build and Publish Docker Image
```
iamgusain, line 5: >name [](http://example.com/codeflow?start=0&length=4)

add a description: as well?

iamgusain (approved):
:shipit:

azure-pipelines.yml
p3dr0rv: done

PR #80, opened 2021-04-29
amgusain/addSpotBugsPlugin

iamgusain:


plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
```diff
@@ -0,0 +1,45 @@
...
+//  THE SOFTWARE.
+
+package com.microsoft.identity.buildsystem;
+import com.github.spotbugs.snom.SpotBugsExtension;
+import com.github.spotbugs.snom.SpotBugsPlugin;
+
+import org.gradle.api.Project;
+import java.io.File;
+
+public final class SpotBugs {
```
shahzaibj, line 31: I think this needs to be a plugin

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
```diff
@@ -0,0 +1,45 @@
...
+    private static final String BASELINE_FILE_PATH_RELATIVE_TO_PROJECT_ROOT = "config/spotbugs/baseline.xml";
+
+    static void apply(Project project) {
+        if(!project.getPlugins().hasPlugin(SpotBugsPlugin.class)) {
+            project.getPlugins().apply(SpotBugsPlugin.class);
+        }
+
+        SpotBugsExtension spotBugsExtension = project.getExtensions().findByType(SpotBugsExtension.class);
```
shoatman, line 39: I'm wondering if this will be part of the existing config section or a completely different section.

shoatman (commented):
Have a question before I approve... want to keep the extension/configuration for the build system together.... if possible.

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/BuildPlugin.java
```diff
@@ -24,6 +24,7 @@ public void apply(final Project project) {
             }
         });
 
+        SpotBugs.apply(project);
```
shahzaibj, line 27:
```suggestion
project.getPlugins().apply(SpotBugs.class);
```

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
```diff
@@ -0,0 +1,45 @@
...
+import com.github.spotbugs.snom.SpotBugsPlugin;
+
+import org.gradle.api.Project;
+import java.io.File;
+
+public final class SpotBugs {
+    private static final String BASELINE_FILE_PATH_RELATIVE_TO_PROJECT_ROOT = "config/spotbugs/baseline.xml";
+
+    static void apply(Project project) {
+        if(!project.getPlugins().hasPlugin(SpotBugsPlugin.class)) {
```
shahzaibj, line 35: I believe this is the official spotbugs plugin?

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
```diff
@@ -0,0 +1,45 @@
...
+            project.getPlugins().apply(SpotBugsPlugin.class);
+        }
+
+        SpotBugsExtension spotBugsExtension = project.getExtensions().findByType(SpotBugsExtension.class);
+        File baselineFile = project.file(BASELINE_FILE_PATH_RELATIVE_TO_PROJECT_ROOT);
```
shahzaibj, line 40: nit: final

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
shahzaibj, line 31: javadoc would be nice

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/BuildPlugin.java
iamgusain: I moved logic to apply and configure the SpotBugs Plugin in its own class SpotBugs.java

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
iamgusain: This is just the class that handles applying and configuring the plugin.. The plugin is actually applied in the apply method of this class

plugins/buildsystem/docs/Spotbugs.md
```diff
@@ -0,0 +1,45 @@
...
+## What is SpotBugs
+
+SpotBugs is a static analysis tool to find bugs in Java programs. It looks for instances of “bug patterns” — code instances that are likely to be errors and reports them as violations. More details can be found at [official docs](https://spotbugs.readthedocs.io/en/stable/introduction.html).
+
+## SpotBugs Configs applied by Auth Client BuildPlugin
+
+Below are the specific values that BuildPlugin sets in addition to the [default values](https://spotbugs-gradle-plugin.netlify.app/com/github/spotbugs/snom/spotbugstask) already set by the SpotBugs Plugin
+
+```groovy
+ spotbugsMain {
```
shahzaibj, line 24: Where is this defined? Didn't see it in this PR

shahzaibj (changes requested):
left some comments

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
shahzaibj: So this class is not a plugin? It seems to write code in a pattern that makes it seem like a plugin...for instance there is an apply method...if it is not a plugin then let's rename the PR / documentation and this method

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
iamgusain: Not sure if I understood correctly, let me try to answer anyways :)

Currently I am keeping it such that spotbugs can be onfigured in its own config section and not from BuildPlugins config.  This is so that we can use the Spotbugs official documentation to learn and  change the config 
something like below

tasks.withType(com.github.spotbugs.snom.SpotBugsTask) {
    baselineFile = file('<PATH TO OVERRIDDEN BASELINE FILE')
    reports {
        xml.enabled = false
        html.enabled = true
    }
}

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
shahzaibj: This PR does not add a custom SpotsBugsPlugin as I understand now...it just applies the official one...so the PR decscription should reflect that....also if `SpotBugs.java` is not a plugin then it shouldn't have an `apply` method that makes it seem like it is one

shahzaibj: And even though you didn't make it a plugin...you actually could make it a plugin...this would become a custom SpotBugs plugin that wraps around the official one

shahzaibj: but that is probably not needed now

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
iamgusain: yes

plugins/buildsystem/docs/Spotbugs.md
iamgusain: updated the doc

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
```diff
@@ -0,0 +1,45 @@
...
+import com.github.spotbugs.snom.SpotBugsExtension;
+import com.github.spotbugs.snom.SpotBugsPlugin;
+
+import org.gradle.api.Project;
+import java.io.File;
+
+public final class SpotBugs {
+    private static final String BASELINE_FILE_PATH_RELATIVE_TO_PROJECT_ROOT = "config/spotbugs/baseline.xml";
+
+    static void apply(Project project) {
```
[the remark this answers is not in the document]
iamgusain: We dont need to make this a plugin, as we are using the official plugin directly

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
iamgusain: We shouldn't make a custom plugin when an official plugin exists and can be used as-is unless we want to customize it. 
The reason we are adding it here in our Custom Build Plugin is so that it can be centrally configured in one place for all our projects, and we don't have to modify same things in individual build.gradle files..

shoatman:
One other comment.. per discussion yesterday on PR description.  Please create one for posterity.  Feel free to point to place in your doc for details.

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
shahzaibj: @iamgusain I totally agree with you....I got the impression that you wanted to make a custom plugin from two things:

- PR title - addSpotBugsPlugin - seems to imply you added a new plugin to the repo for SpotBugs (but you didn't)
- The code design pattern inside `SpotBugs.java` - this class is written following the same pattern as we would when creating a plugin i.e there is an apply method with the same signature as the official apply method of a plugin...so when I reviewed this code I thought you were trying to create a plugin but accidently forgot to implement the interface and accidentally made the method a static...but you didn't...you were doing it on purpose...so my recommendation would be to just change the method name from `apply` to something else to avoid confusion. (You actually probably didn't even need a class for `SpotBugs.java` - the code here could just go directly in the existing plugin class, but I would leave that up to you...but we should definitely rename the method at least.)

shahzaibj: I think even the class needs to be renamed...what you really made here is a `SpotBugsApplicator`

iamgusain:
> Have a question before I approve... want to keep the extension/configuration for the build system together.... if possible.

I would love to have it controlled by one single config. But as we are applying the official plugin directly it can be configured separately as well. I am not sure if there is a way to prevent that.  
But honestly I am not expecting us to override the default configurations for this plugin much for individual projects. So we might not have to worry about that.

iamgusain:
> One other comment.. per discussion yesterday on PR description. Please create one for posterity. Feel free to point to place in your doc for details.

The github desktop description didnt auto apply when I created the PR :(. I have updated the description now. Thanks for pointing

plugins/buildsystem/changelog
```diff
@@ -0,0 +1,13 @@
+Plugin Overview: https://github.com/AzureAD/android-complete/blob/master/plugins/buildsystem/docs/Overview.md
```
iamgusain, line 1: @shoatman Is this the right way to enable changelogs? (by adding changelog file under plugins/buildsystem)

shoatman: That looks right.  Thanks

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
iamgusain: As you mentioned to make it a plugin it needs to implement the Plugin<T>, which will force it the override apply method. Without that its just a normal class. 

I am keeping it in a separate class and not with the BuildPlugin class for two reasons
1. All the spotbugs related logic can live here. 
2. The BuildPlugin class remains compact and easily readable what its doing. (Think of if from future perspective if we add more such plugins, it will be better to keep them in separate classes, rather than putting all the logic in one single file )

I think the name of the Spotbugs is fine, as it contains logic to apply Spotbugs plugin as well as other spotbugs related configs. Let me know if you still think otherwise.. happy to discuss offline.

plugins/buildsystem/src/test/java/com/microsoft/identity/buildsystem/BuildPluginTests.java
```diff
@@ -0,0 +1,55 @@
...
+    public void doesNotApplyDeSugaringByDefaultToAndroidProjects() {
+        Project project = ProjectBuilder.builder().build();
+        project.getPluginManager().apply("com.android.library");
+        project.getPluginManager().apply("com.microsoft.identity.buildsystem");
+
+        LibraryExtension libraryExtension = project.getExtensions().findByType(LibraryExtension.class);
+        assertFalse(libraryExtension.getCompileOptions().isCoreLibraryDesugaringEnabled());
+    }
+
+}
```
shahzaibj, line 55: new line

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
shahzaibj: My primary concern is around the method name and its signature. Generally the `apply(Project project)` method is a method that lives in a plugin and used for applying that plugin. 

Over here we've made a method with same name and signature and put it in a class that lives in a plugin project except that the class itself isn't actually a plugin. I think it is just that at an initial glace it seems confusing and to the reader it seems you've tried to create a plugin when it is not. (I was confused and hence requested changes on this PR). I think we could really call this method anything (may be `applySpotBugsPlugin`) except `apply` to make that distinction. 

With that said, I don't think there is any perfect way to do this so I'm going to let you pick..

iamgusain: I think that's a valid point. Updated the method name to `applySpotBugsPlugin`

plugins/buildsystem/build.gradle
```diff
@@ -28,12 +28,16 @@ gradlePlugin {
 repositories {
     mavenCentral()
     google()
+    maven {
+        url "https://plugins.gradle.org/m2/"
+    }
```
shahzaibj, line 33: I believe this URL just corresponds to the gradle plugin portal

```suggestion
    gradlePluginPortal()
```

plugins/buildsystem/build.gradle
```diff
@@ -28,12 +28,16 @@ gradlePlugin {
...
+    maven {
+        url "https://plugins.gradle.org/m2/"
+    }
 }
 
 dependencies {
     testImplementation 'org.junit.jupiter:junit-jupiter-api:5.6.0'
     testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
     implementation 'com.android.tools.build:gradle:4.1.0'
+    implementation 'gradle.plugin.com.github.spotbugs.snom:spotbugs-gradle-plugin:4.7.1'
```
shahzaibj, line 40: Can we move these versions to a version file?

shahzaibj (approved):

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/SpotBugs.java
shoatman: k...thanks

plugins/buildsystem/build.gradle
iamgusain: Lets do this as part of this work item: https://identitydivision.visualstudio.com/Engineering/_backlogs/backlog/Auth%20Client%20-%20Android/Features/?workitem=1359176

AdamBJohnsonx (approved):

PR #83, opened 2021-04-30
Reference common java as common4j

shahzaibj:
Better to use the same name as what it is actually called in the directory structure...`common/common4j` so let's just call it `common4j`


iamgusain (approved):
:shipit:

settings.gradle
```diff
@@ -19,15 +19,15 @@ pluginManagement {
...
         ':labapi', ':keyvault', ':testutils', 'msalautomationapp', 'uiautomationutilities', 'brokerautomationapp',
         ':oneauth', ':oneauthtestapp', ':msalcpptesthost', ':AzureSample', 'AcaPlugin'
         //, ':tsl', ':AzureSample',
 project(':adal').projectDir = new File('adal/adal')
 project(':msal').projectDir = new File('msal/msal')
 project(':common').projectDir = new File('common/common')
-project(':common-java').projectDir = new File('common/common4j')
+project(':common4j').projectDir = new File('common/common4j')
 project(':AADAuthenticator').projectDir = new File('broker/AADAuthenticator')
 project(':brokerHost').projectDir = new File('broker/userapp')
```
iamgusain, line 32: >brokerHost [](http://example.com/codeflow?start=10&length=10)

on the similar lines. should this be renamed to brokerUserApp?

shahzaibj: It can be but that would require changes in a lot of places (primarily pipelines) so I am going to leave as is for now.

PR #86, opened 2021-05-04
Add droid command to setup our devex projects only

shahzaibj:
This is per discussion here: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1342/files#r624265379

cc/ @p3dr0rv @iamgusain


.gitconfig
```diff
@@ -19,6 +19,16 @@
 		workDir=$(pwd); \
 		git config --replace-all androidcomplete.workDir $workDir; \
 		}; f"
+	droidSetupDevexOnly = "!f() { \
```
iamgusain, line 22: >droidSetupDevexOnly [](http://example.com/codeflow?start=1&length=19)

when is this supposed to be used as compared to droidSetup, Should we get rid of droidSetup,

Also update the documentation/readme.

iamgusain:
how is this related to  [https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1342/files#r624265379](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1342/files#r624265379)?

shahzaibj:
> how is this related to https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1342/files#r624265379?

@iamgusain This change is required for that pipeline to run:

image

because that pipeline calls `droidSetupDevexOnly` to avoid having to figure out how to clone private repos (ideally we should figure that out too but in interest of time I've done this.)

.gitconfig
shahzaibj: Folks should not be using this new command because it won't help them much as it doesn't clone all the repos they need. It is sort of useless unless you are fine with a partially working Android-Complete. I made this primarily for the pipeline to be able to build AzureSample app with local variant MSAL because this is the only way to do it as of now.

and we should not get rid of `droidSetup` because everyone is using that and should be using that.

shahzaibj: This is basically I work-around for now until we figure out how to clone private repos in the pipeline....then we can just use droidSetup in the pipeline too...

p3dr0rv (approved):
LGTM

PR #104, opened 2021-07-06
BrokerHost pipeline from Android-Complete

p3dr0rv:
Pipeline that build brokerHost from android-complete, if we try to build the local version from ad-accounts this will result in a misleading build that is using dist dependencies  

    if (findProject(':adal') != null) {
        localApi(project(':adal')) {
            exclude module: 'common'
        }
        distApi(group: 'com.microsoft.aad', name: 'adal', version: "${adalDistVersion}") {
            exclude module: 'common'
        }
    } else {
        api(group: 'com.microsoft.aad', name: 'adal', version: "${adalDistVersion}") {
            exclude module: 'common'
        }
    }

Pipeline: https://identitydivision.visualstudio.com/Engineering/_build?definitionId=1432&_a=summary

### DIFF
--- a/azure-pipelines/build-android-complete.yml
+++ b/azure-pipelines/build-android-complete.yml
@@ -1,0 +1,47 @@
+# File: azure-pipelines\pull-request-validation\common4j.yml
+# Description: Run test in common4j
+name: $(date:yyyyMMdd)$(rev:.r)
+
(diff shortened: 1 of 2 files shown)


shahzaibj (approved):
This is excellent 💯

PR #105, opened 2021-07-15
Add a java-complete root project to android-complete repo

shahzaibj:
Video explaining what's going on: [Video](https://microsoft-my.sharepoint-df.com/:v:/p/shjameel/EajZkLc162VEqH8GrX2apKAB0i4ghDKYhgoMZSOjP02wPw?e=hK4Wl3)

Buildscript dependencies are applied to all modules in a project, and for us we apply the [Android Build Tools to all of them](https://github.com/AzureAD/android-complete/blob/master/build.gradle) because historically we've only had Android modules, however, now that we have Pure Java modules we must build them without the presence of Android Build Tools on the classpath. To achieve this, I've added a new root project called as `java-complete` and this root project only consumes the Pure Java modules as part of its `settings.gradle` and within the buildscript of this project, we DON'T take a depedency on the android build tools. This allows the build to happen without their presence. 

Our older root project, which was present on the root of git repo, should continue to function as it did before and essentially that acts as an "Android" root project that consumes both our Pure Java modules as well as Android modules.

Related common PR: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1468
Related broker PR: https://github.com/AzureAD/ad-accounts-for-android/pull/1625


java-complete/gradle/versions.gradle
```diff
@@ -0,0 +1,85 @@
+// Variables for entire project
+ext {
+    // SDK
+    minSdkVersion = 16
```
AdamBJohnsonx, line 4: Is this still the right version?

java-complete/gradle/versions.gradle
```diff
@@ -0,0 +1,85 @@
...
+    automationAppMinSDKVersion = 21
+    targetSdkVersion = 30
+    compileSdkVersion = 30
+    buildToolsVersion = "28.0.3"
+
+    // Plugins
+    gradleVersion = '4.1.0'
```
AdamBJohnsonx, line 11: 4.2.2?

java-complete/gradle/versions.gradle
```diff
@@ -0,0 +1,85 @@
+// Variables for entire project
```
AdamBJohnsonx, line 1: Is it possible to tree these dependency files at all, so that we don't need to include everything from everywhere in each of them?

AdamBJohnsonx (approved):
minor comments

java-complete/gradle/versions.gradle
shahzaibj: I've actually deleted the entire version file as I realized I could just use the same one located at repo root. Done here: https://github.com/AzureAD/android-complete/pull/105/commits/dccfa95e1abd228af747aca4ab0c06d469f18c24

java-complete/gradle/versions.gradle
shahzaibj: Addressed per: https://github.com/AzureAD/android-complete/pull/105#discussion_r671579801

java-complete/gradle/versions.gradle
shahzaibj: Addressed by deleting the version file as per: https://github.com/AzureAD/android-complete/pull/105#discussion_r671579801

PR #109, opened 2021-08-03
Upgrade gradle versions to 7.0

AdamBJohnsonx:


AdamBJohnsonx:
We can't have this because our build agents would need to run java 11.

PR #110, opened 2021-08-30
Code Coverage Plugin

paulkagiri:
# Code Coverage Plugin 

## Intro
- [Code coverage](https://en.wikipedia.org/wiki/Code_coverage) is a software metric used to measure how many lines of our code are executed during automated tests.

- [JaCoCo](https://www.eclemma.org/jacoco/trunk/index.html) is a free Java code coverage tool and the [JaCoCo plugin](https://docs.gradle.org/current/userguide/jacoco_plugin.html) provides code coverage metrics for Java code via integration with JaCoCo.

## Why
In order to generate [JaCoCo](https://www.jacoco.org/jacoco/trunk/doc/index.html) unit test coverage reports for Android projects you need to create `JacocoReport` tasks and configure them by providing paths to source code, execution data and compiled classes. It's not straightforward since Android projects can have different flavors and build types thus requiring additional paths to be set. This plugin configures these `JacocoReport` tasks automatically.

## Usage
```groovy
plugins {
    ...
    id 'com.gradle.plugin-publish' version '0.14.0' // or whatever version is most recent
}

codeCoverageReport{
    html.enabled = true // by default it's true
    xml.enabled = true // by default it's true
    csv.enabled = true // by default it's false

    unitTests.enabled = true // whether code coverage tasks for unit tests will be generated
    androidTests.enabled = true // whether code coverage tasks for instrumentation tests will be generated

    excludeFlavors = [''] // the product flavors to exclude when generating the code coverage tasks

    excludeClasses = [''] // additional classes to exclude - most are already catered for

    destination = '/some/other/directory' // if you want to configure a custom path to save the code coverage reports, by default your report gets saved in `[project]/build/jacoco/{flavor}{build type}{project}{test type}CoverageReport`

    includeNoLocationClasses = true // To include Robolectric tests in the Jacoco report this needs to be true
}

android {
    buildTypes {
        debug {
            testCoverageEnabled true // this instructs the plugin to generate code coverage reports for this build type
            ...
        }
        release {
            testCoverageEnabled false // this instructs the plugin to NOT generate code coverage reports for this build type
            ...
        }
    }
    ...
    productFlavors {
        local {}
        dist {}
    }
}
```

The above configuration creates a `JacocoReport` task for each variant in the form of `{flavor}{build type}{project}{test type}CoverageReport`
```
distDebugAppAndroidTestCoverageReport
distDebugAppUnitTestCoverageReport
localDebugAppAndroidTestCoverageReport
localDebugAppUnitTestCoverageReport
```

By default these are the excluded classes under [Constants](https://github.com/AzureAD/android-complete/blob/paul/code-coverage-plugin/plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/Constants.kt#L25)


plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
```diff
@@ -0,0 +1,174 @@
...
+            }
+        }
+
+        private fun getFileFilterPatterns(): List<String> = JacocoCoverageReportExtension.DEFAULT_EXCLUDES + reportExtension.excludes
+
+        private fun Project.fileTree(dir: Any, excludes: List<String> = listOf(), includes: List<String> = listOf()): ConfigurableFileTree =
+                fileTree(mapOf("dir" to dir, "excludes" to excludes, "includes" to includes))
+
+    }
+}
```
AdamBJohnsonx, line 174:
```suggestion
}

```

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
```diff
@@ -0,0 +1,174 @@
...
+import org.gradle.api.logging.Logger
+import org.gradle.api.logging.Logging
+import org.gradle.api.plugins.PluginContainer
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.testing.jacoco.plugins.JacocoPlugin
+import org.gradle.testing.jacoco.plugins.JacocoTaskExtension
+import org.gradle.testing.jacoco.tasks.JacocoReport
+import java.io.File
+
+class CodeCoveragePlugin {
```
AdamBJohnsonx, line 38: We should have the same standards for our build tools as for our code.  Documentation for classes, etc.

AdamBJohnsonx:
We should have a description indicating what this does, precisely, in the PR description, and links to the documentation of the features used in the code comments, so future authors can have a starting place to go when looking at this code.

.gitignore
```diff
@@ -43,4 +43,6 @@ ehthumbs.db
 ehthumbs_vista.db
 
 # Folder config file
-[Dd]esktop.ini
\ No newline at end of file
+[Dd]esktop.ini
+
+out/
```
shahzaibj, line 48: nit: new line ending

plugins/buildsystem/build.gradle
```diff
@@ -31,11 +33,19 @@ repositories {
     gradlePluginPortal()
 }
 
+ext {
+    kotlinVersion = '1.5.30'
```
shahzaibj, line 37: probably define it in the versions.gradle file

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
```diff
@@ -0,0 +1,261 @@
...
+import org.gradle.api.tasks.testing.Test
+import org.gradle.testing.jacoco.plugins.JacocoPlugin
+import org.gradle.testing.jacoco.plugins.JacocoTaskExtension
+import org.gradle.testing.jacoco.tasks.JacocoReport
+import java.io.File
+
+/**
+ * This class creates code coverage tasks in the given project
+ */
+object CodeCoveragePlugin {
```
shahzaibj, line 46: This isn't a plugin in and of itself and therefore we should be careful with naming. This to me just seems a class that applies a different jacoco plugin.

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
```diff
@@ -0,0 +1,261 @@
...
+
+    private lateinit var reportExtension: CodeCoverageReportExtension
```
shahzaibj, line 48: For those of us that don't know `kotlin`, I think we need a brown bag on this. Like I don't know what `lateinit` does..and what is the reason for using that here?

shahzaibj:
An overall comment, was there a specific reason for using kotlin here? 

I mean its nice, but I think we also need to get the team onboarded with kotlin as well.

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
```diff
@@ -0,0 +1,261 @@
...
+    /**
+     * Gets the codeCoverageReport configurations and uses them to create the code coverage tasks
+     */
+    @JvmStatic
+    fun applyCodeCoveragePlugin(project: Project) {
+        // get the configurations under codeCoverageReport
+        reportExtension = project.extensions.create("codeCoverageReport", CodeCoverageReportExtension::class.java)
+
+        // apply jacoco
+        if (project.plugins.withType(JacocoPlugin::class.java).isEmpty()) {
```
shahzaibj, line 59: I think we need to know what this Jacoco plugin is...it doesn't seem to be ours. Can you document this somewhere? probably in the doc you created but would be nice to also leave a pointer here in the code

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
```diff
@@ -0,0 +1,261 @@
...
+    /**
+     * Check if we are allowed to create code coverage tasks for this variant
+     */
+    private fun shouldCreateTaskForVariant(excludeFlavours: List<String>, variant: BaseVariant, testType: String): Boolean {
+        if (!variant.buildType.isTestCoverageEnabled || excludeFlavours.contains(variant.flavorName.toLowerCase())) {
+            return false
+        }
+
+        // AFAIK, the android tests tasks are only generated for debug build types
+        if (testType == TestTypes.AndroidTest && variant.buildType.name != "debug") {
```
shahzaibj, line 126: It is generated for one build type only...usually it is `debug` but developers may configure it to be something else. For now I think we can assume `debug` as our projects are using debug

shahzaibj:
How would this plugin work on Non-Android gradle modules such as `common4j` or `broker4j`? Have we tested it?

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
AdamBJohnsonx: There was one scheduled at one point and put off for a long time.  This one is tricky - Kotlin doesn't allow null values, but lateinit variables are allowed to be null without null checks because it is assumed that they are guaranteed to be supplied before use.  Is that necessary here?

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
AdamBJohnsonx: Seems to be part of gradle itself? https://docs.gradle.org/current/javadoc/org/gradle/testing/jacoco/plugins/JacocoPlugin.html

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
paulkagiri: I agree with this

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
paulkagiri: Yeah it's required, otherwise the variable `reportExtension` would have to be nullable

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
paulkagiri: I agree

plugins/buildsystem/build.gradle
paulkagiri: The `buildsystem` plugin is a separate project and does not use the android-complete project configurations, will speak to @shoatman about it

paulkagiri:
> An overall comment, was there a specific reason for using kotlin here?
> 
> I mean its nice, but I think we also need to get the team onboarded with kotlin as well.

I agree with you, will arrange a brownbag on Kotlin

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoveragePlugin.kt
paulkagiri: Updated the description

paulkagiri:
> How would this plugin work on Non-Android gradle modules such as `common4j` or `broker4j`? Have we tested it?

Non-Android projects support is being added

paulkagiri:
> How would this plugin work on Non-Android gradle modules such as `common4j` or `broker4j`? Have we tested it?

The support for Java projects is now available

plugins/buildsystem/build.gradle
AdamBJohnsonx: Is this still a concern?

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
```diff
@@ -0,0 +1,200 @@
...
+            toolVersion = reportExtension.jacocoVersion
+        }
+
+        tasks.withType(Test::class.java) { testTask ->
+            testTask.extensions.findByType(JacocoTaskExtension::class.java)?.apply {
+                // To include Robolectric tests in the Jacoco report, flag -> "includeNolocationClasses" should be set to true
+                isIncludeNoLocationClasses = reportExtension.includeNoLocationClasses
+                if (isIncludeNoLocationClasses) {
+                    // This needs to be excluded for JDK 11
+                    // SEE: https://support.circleci.com/hc/en-us/articles/360047926852-Android-Builds-Fail-with-java-lang-ClassNotFoundException-jdk-internal-reflect-GeneratedSerializationConstructorAccessor1-
```
AdamBJohnsonx, line 78: This link doesn't work for me.  Maybe https://github.com/gradle/gradle/issues/5184 ?

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
```diff
@@ -0,0 +1,200 @@
...
+
+        if (reportExtension.androidTests.enabled) {
+            createTask(project, TestTypes.AndroidTest)
+        }
+    }
+
+    /**
+     * Creates the code coverage tasks for the different build variants
+     */
+    private fun createTask(project: Project, testType: String) {
```
AdamBJohnsonx, line 103: nit: createTasks?  This seems to create and add multiples.

.gitignore
AdamBJohnsonx, line 48: do we need this checked in?

AdamBJohnsonx (approved):
Minor comments.  If it works, I'm for it.

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
paulkagiri: Done

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
paulkagiri: Updated

plugins/buildsystem/build.gradle
paulkagiri: This has been updated

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverageReportExtension.kt
```diff
@@ -0,0 +1,55 @@
...
+//  IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+//  FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+//  AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+//  LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+//  OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+//  THE SOFTWARE.
+package com.microsoft.identity.buildsystem.codecov
+
+/**
+ * This class controls some of the configurations used by the code coverage plugin
```
shahzaibj, line 26:
```suggestion
 * This class controls some of the configurations used by the code coverage plugin.
```

shahzaibj:
Were we able to verify that just running the tests doesn't run coverage? (basically only code coverage specific gradle tasks will run the coverage)

shahzaibj (approved):
It looks fine to me. I think we just need to see it working e2e now.

paulkagiri:
Cool, I will go ahead and merge this in, thanks @AdamBJohnsonx & @shahzaibj

paulkagiri:
> Were we able to verify that just running the tests doesn't run coverage? (basically only code coverage specific gradle tasks will run the coverage)

Yeah, these changes made that possible - https://github.com/AzureAD/android-complete/pull/110/commits/c1145c5c63fc8911a4d3446b2c471bd35b4c078c

PR #115, opened 2021-10-14
Fix for the instrumented tests

paulkagiri:
- This fixes issues caused by the first version of the code coverage tool specifically in instrumented tests
- The issue is that jacoco is not initialized properly when running the instrumented tests
- The fix follows this -https://github.com/jacoco/jacoco/issues/968

### DIFF
--- a/plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
+++ b/plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
@@ -93,6 +93,8 @@ object CodeCoverage {
         }
 
         if (reportExtension.androidTests.enabled) {
+            // need to create below file so that jacoco is properly initialized during instrumented tests
+            // see https://github.com/jacoco/jacoco/issues/968 for more details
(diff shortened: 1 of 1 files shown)


plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
```diff
@@ -93,6 +93,13 @@ object CodeCoverage {
         }
 
         if (reportExtension.androidTests.enabled) {
+            val androidTestResourcesDirectory = project.file("src/androidTest/resources")
+            val jacocoAgentProperties = File(androidTestResourcesDirectory, "jacoco-agent.properties")
+            if (!jacocoAgentProperties.exists()) {
+                androidTestResourcesDirectory.mkdirs()
+                jacocoAgentProperties.writeText("output=none")
```
AdamBJohnsonx, line 100: Do we want no output?

AdamBJohnsonx: Could you add a comment in the code with a link for things that are counterintuitive.  That way somebody else doesn't have to figure it out later.

PR #118, opened 2021-11-03
Jacoco initialization issue fix

paulkagiri:
- This fixes issues caused by the first version of the code coverage tool specifically in instrumented tests
- The issue is that jacoco is not initialized properly when running the some unit tests
- The fix follows this -https://github.com/jacoco/jacoco/issues/968

### DIFF
--- a/plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
+++ b/plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
@@ -63,6 +63,17 @@ object CodeCoverage {
      * Apply some plugin configuration from [CodeCoverageReportExtension] to the project.
      */
     private fun Project.configure() {
+        if (isAndroidProject(this)) {
+            // need to create below file so that jacoco is properly initialized during instrumented tests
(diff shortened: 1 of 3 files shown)


AdamBJohnsonx (approved):
I don't have a problem here... but we could also just check in a copy of the file.

paulkagiri:
> I don't have a problem here... but we could also just check in a copy of the file.

It will be generated if it does not exist

plugins/buildsystem/src/main/java/com/microsoft/identity/buildsystem/codecov/CodeCoverage.kt
```diff
@@ -63,6 +63,17 @@ object CodeCoverage {
      * Apply some plugin configuration from [CodeCoverageReportExtension] to the project.
      */
     private fun Project.configure() {
+        if (isAndroidProject(this)) {
+            // need to create below file so that jacoco is properly initialized during instrumented tests
+            // see https://github.com/jacoco/jacoco/issues/968 for more details
+            val androidTestResourcesDirectory = project.file("src/androidTest/resources")
```
TWiStErRob, line 69: please note https://github.com/jacoco/jacoco/issues/968#issuecomment-969016476

PR #122, opened 2021-12-17
Add Continuous Delivery for LinuxBrokerPackager & Java Linux Broker Test App

wainaina:
Adds:
 - Linux Broker Package project
 - Java Linux Broker Test App project 
to the continuous delivery pipeline.

The packages are assembled, and packaged using <PROJECTNAME>:buildDeb, 

Then the packages are published as pipeline artifacts that can then be consumed by the publishing pipeline for publishing on packages.microsoft.com
Pipeline link: https://identitydivision.visualstudio.com/DefaultCollection/Engineering/_build?definitionId=1515

Deb publisher link: https://identitydivision.visualstudio.com/IDDP/_build?definitionId=1447&_a=summary

Dependent on the approval of: https://github.com/AzureAD/ad-accounts-for-android/pull/1797


azure-pipelines/continuous-delivery/auth-client-android-dev.yml
```diff
@@ -108,6 +109,53 @@ stages:
...
+          env:
+            ENV_VSTS_MVN_ANDROIDADACCOUNTS_USERNAME: $(mvnUserName)
+            ENV_VSTS_MVN_ANDROIDADACCOUNTS_ACCESSTOKEN: $(mvnAccessToken)
+          inputs:
+            cwd: $(Build.SourcesDirectory)/broker-java-root
+            tasks: LinuxBrokerPackage:clean LinuxBrokerPackage:buildDeb $(projVersionParam) $(linuxBrokerVersion) --build-cache --info
+            publishJUnitResults: false
+            jdkArchitecture: x86
+            sqAnalysisBreakBuildIfQualityGateFailed: false
+        - task: CopyFiles@2
```
iamgusain, line 123: We can use this pipeline to publish directly to packages.micrsoft.com, instead of publishing artifacts here and then using another pipeline to do the actual publish.

You can create a template for publishing to packages.microsoft.com and then use that template in this file.

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
```diff
@@ -108,6 +109,53 @@ stages:
...
+            SourceFolder: LinuxBrokerPackage/build/distributions
+            TargetFolder: $(build.artifactstagingdirectory)
+        - task: PublishPipelineArtifact@1
+          name: PublishPipelineArtifacts1
+          displayName: 'Publish Artifact: LinuxBrokerPackage Linux Test App Deb'
+          inputs:
+            ArtifactName: LinuxBrokerPackage Deb Package
+            TargetPath: $(build.artifactstagingdirectory)
+        - task: Gradle@3 
+          displayName: Build and publish java linux broker test app
```
iamgusain, line 136: I was thinking using this pipeline only for our production libraries consumed by our partners/customers and create a separate pipeline for all the test apps. 
Can you create a separate pipeline auth-client-android-testapps.yml and move it there?

iamgusain (changes requested):

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
wainaina: Unless we publish to a different repo on packages.microsoft.com. Currently, the publishing pipelines point to prod repos.

shahzaibj: We should actually not be publishing to `packages.microsoft.com` from continous delivery pipeline. My understanding is that this runs daily? For publishing to packages, we would want to be more explicit about when we publish and probably should have a dedicated pipeline that runs at-will.

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
wainaina: Removing the Java Linux Test App.

versioning/version.properties
```diff
@@ -0,0 +1,4 @@
+#Thu Sep 02 18:50:06 PDT 2021
```
iamgusain, line 1: Why do we need this and the version_tasks.gradle file in android-complete?

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
iamgusain: Does that mean we don't need to publish the deb daily, in that case we probably don't need this task here.

shahzaibj: We want a deb daily but only as a pipeline artifact (we don't want that to get pushed to `packages.microsoft.com`)

versioning/version.properties
wainaina: Removed.

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
wainaina: We shall have 2 publishing events:
1. Publishing to the pipeline. - Only one in this pipeline
2. Publishing to packages.microsoft.com
Candidate builds that are chosen to be published are ran through another pipeline with more tasks to publish these packages to the artifact repo.

shoatman (approved):

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
iamgusain: Is the plan to consume the pipeline artifact on daily/ongoing basis to do some validations (automation/ self host/dogfood) or only when we need to publish it to pkg.msft.com.

shahzaibj: @iamgusain Yes, we would use these for automation and self host.

iamgusain (approved):

PR #125, opened 2022-02-10
Do not set buildDir in android-complete and java-complete build.gradle

fadidurah:
The default location for the build folder is projectDir/build. Originally, we were building everything to C:/temp because of windows long path restrictions, but those can now be addressed with registry settings.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -31,9 +31,9 @@ buildscript {
 }
 
 allprojects {
-    if (org.gradle.internal.os.OperatingSystem.current().isWindows()) {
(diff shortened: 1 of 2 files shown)


p3dr0rv:
can we remove this from java-complete too?
@shahzaibj

iamgusain (approved):

shahzaibj:
> can we remove this from java-complete too? @shahzaibj

Yes, I think that should be fine.

fadidurah:
I will remove from java-complete also

p3dr0rv (approved):

PR #132, opened 2022-04-11
Updated plugin

shoatman:


shoatman:
Added support for plugin used by OneAuth.  removed pop benchmarker since not frequently used.

PR #139, opened 2022-05-20
Update broker-test.yml for Azure Pipelines

p3dr0rv:


shahzaibj:
Can you also change firebase deviceId to `blueline` and API version to `28` because `flame` device is no longer in firebase

shahzaibj (approved):

PR #140, opened 2022-05-23
Add microsoft diagnostics version & project details to android-complete and assoc. projects

wainaina:


install-from-source.sh
```diff
@@ -0,0 +1,183 @@
+#!/bin/sh
```
shahzaibj, line 1: What is this entire script for?

wainaina: I should remove this script. This script is autogenerated by the install task.

PR #145, opened 2022-07-26
Use one maven access token through all the test app pipelines.

p3dr0rv:
This is part of an effort to remove all the copies of maven access token in each pipeline and have just one token, so we don't have to update multiple pipelines.

Library to store VSTS access tokens: 
https://identitydivision.visualstudio.com/Engineering/_library?itemType=VariableGroups&view=VariableGroupView&variableGroupId=14&path=AndroidAuthClientVariables

Also remove variables in branches in favor to use resources tab.

### Results:
https://identitydivision.visualstudio.com/Engineering/_build?definitionScope=%5CAndroid%5CTesting%20and%20Automation%20Pipelines%5CTest%20Apps

### DIFF
--- a/azure-pipelines/test-app/adal-test-app.yml
+++ b/azure-pipelines/test-app/adal-test-app.yml
@@ -2,10 +2,8 @@
 # Description: Build ADAL test app APK
 # use local_<project>_branch variables to build the Local version
 # use dist_<project>_version variables to build the Dist version
-# Variable 'local_[adal, common]_branch' was defined in the Variables tab
 # Variable 'dist_[adal, common]_version' was defined in the Variables tab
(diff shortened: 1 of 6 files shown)


azure-pipelines/test-app/adal-test-app.yml
```diff
@@ -40,19 +38,21 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: $(local_common_branch)
```
iamgusain, line 43: Did you also remove these from the variables in the pipeline in ADO?

iamgusain (approved):

azure-pipelines/test-app/adal-test-app.yml
p3dr0rv: I plan to do it once this is merged

PR #147, opened 2022-08-03
Remove references to the OneAuthAndroid feed in upstreamsources.md

remyjette:
OneAuth does not use this feed. Updated this doc to reflect what we actually
use.


p3dr0rv (approved):
LGTM

PR #151, opened 2022-08-24
enable auto-google-login on test devices

fadidurah:
Quick PR to add the auto-google-login arguement to broker CI pipeline, allowing runner to login to a google account on the physical testing device to access play store and other features.

### DIFF
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -229,6 +229,7 @@ stages:
             --type instrumentation \
             --app "$(Pipeline.Workspace)/msalautomationapks/$(msalApp)" \
             --test "$(Pipeline.Workspace)/msalautomationapks/$(msaltestApp)" \
+            --auto-google-login \
             --record-video \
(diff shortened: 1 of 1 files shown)


azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -229,6 +229,7 @@ stages:
             --type instrumentation \
             --app "$(Pipeline.Workspace)/msalautomationapks/$(msalApp)" \
             --test "$(Pipeline.Workspace)/msalautomationapks/$(msaltestApp)" \
+            --auto-google-login \
```
shahzaibj, line 232: Wasn't this already happening? I thought this was a default?

fadidurah: Yes it is set to true by default when using gcloud, but flank actually sets it to false by default, same thing with record-video

shahzaibj (approved):
I'm approving this, but I hope we will leverage the template that you created earlier for these pipelines as well.

fadidurah:
> I'm approving this, but I hope we will leverage the template that you created earlier for these pipelines as well.

I'm thinking I will implement the template structure sometime late september if my other items permit, It's not really blocking anything at the moment and I think my other items are higher priority.

PR #158, opened 2022-10-06
Fix failing test-app pipelines.

gitaumoses4:
The TestApp pipelines are failing https://identitydivision.visualstudio.com/Engineering/_build?definitionScope=%5CAndroid%5CTesting%20and%20Automation%20Pipelines%5CTest%20Apps 

This was a result of this change: https://github.com/AzureAD/android-complete/pull/148/files

We need to read the vstsUsername and vstsAccessToken directly from the environment in both `build.gradle` and `settings.gradle`

### DIFF
--- a/azure-pipelines/test-app/adal-test-app.yml
+++ b/azure-pipelines/test-app/adal-test-app.yml
@@ -84,7 +84,7 @@ jobs:
     displayName: Set MVN AccessToken in Environment
     inputs:
       filename: echo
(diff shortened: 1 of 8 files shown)


shahzaibj (approved):

bktowett (approved):

fadidurah (approved):
Thanks for fixing this, this will unblock some things on automation side as well!

PR #159, opened 2022-10-10
Revert Azure Sample APK Name Change

fadidurah:
Test app pipelines were recently fixed, which caused the azure sample apk name to return to `AzureSample-local-debug.apk`. For a while it was `AzureSample-external-release.apk` due to a band-aid fix we did to get test app pipelines temporarily working.


shahzaibj (approved):

p3dr0rv:
I wonder if we should move these constants to pipeline variables, to avoid making a pr every time an APK name change.

fadidurah:
> I wonder if we should move these constants to pipeline variables, to avoid making a pr every time an APK name change.

Working on this now, just wanted to push this out for the quick fix.

PR #161, opened 2022-10-20
Broker CI Pipeline: Support using multiple devices (multiple API levels), download first party Apps, run some tests on API 29- only

fadidurah:
This work is part of an effort to streamline how tests are ran on our Broker CI Pipeline and increase reliability of pipeline runs. Utilize changes from Common PR to ignore tests that are unreliable on pipeline, as well as selecting tests that are marked to be ran on API29-. Also, download and push first party app apk onto test devices to allow test cases to install them from local memory. Also, it seems that API 32 device is the most consistent with regards to avoiding the network issues that have been observed in firebase devices. API 29- devices are still susceptible to network issues, but they are running much fewer tests and are isolated in their own pipeline stages, so they can be individually re-ran if an engineer sees that a device with a bad network has been provided by firebase. This lets the engineer avoid rerunning the entire test suite. Since the number of tests has been reduced on individual stages, we are not using flank for API29- devices, and reducing test shards on the API30+ devices from 3 to 2. Update how the OldAuthenticator apk is downloaded.

It's probably best to look at this PR after looking at the Common PR, but before MSAL and Broker.
Common: https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/1875
MSAL: https://github.com/AzureAD/microsoft-authentication-library-for-android/pull/1712
Broker: https://github.com/AzureAD/ad-accounts-for-android/pull/2031


shahzaibj (approved):

wainaina (approved):

azure-pipelines/ui-automation/templates/download-first-party-apps.yml
```diff
@@ -0,0 +1,63 @@
...
+        inputs:
+          command: 'download'
+          downloadDirectory: '$(Build.ArtifactStagingDirectory)/firstparty'
+          feedsToUse: 'internal'
+          vstsFeed: '${{ parameters.internalFeedName }}'
+          vstsFeedPackage: '${{ parameters.wordPackage }}'
+          vstsPackageVersion: '${{ parameters.wordVersion }}'
+      - publish: $(Build.ArtifactStagingDirectory)/firstparty
+        displayName: 'Publish First Party APKs'
+        artifact: firstpartyapks
```
wainaina, line 63: nit: Add new line

PR #169, opened 2022-12-20
Download Edge from universal packages, extra stage to run brokerhost tests in broker CI

fadidurah:
Add a step to download edge apk from universal packages. This lets us use LocalApk installation during our UI automation. Also, create an extra stage to run brokerhost tests in Broker CI Pipeline https://dev.azure.com/IdentityDivision/Engineering/_build?definitionId=1490&_a=summary

### DIFF
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -141,6 +141,12 @@ stages:
   - brokers_azure_sample
   displayName: Running MSAL with Broker Test Plan (API ${{ parameters.firebaseDeviceAndroidVersionHigh }})
   jobs:
+    - task: DownloadSecureFile@1
+        displayName: 'Download Certificate File'
+        name: certificateFile
+        inputs:
+          secureFile: mfatest.pfx
+          retryCount: 5
     - template: ./templates/flank/run-on-firebase-with-flank.yml
       parameters:
         automationAppApkPath: "$(Pipeline.Workspace)/msalautomationapks/$(msalApp)"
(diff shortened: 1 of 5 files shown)


azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -102,6 +105,17 @@ stages:
         msalFlavor: Dist
         brokerSource: LocalApk
         brokerUpdateSource: LocalApk
+# msalautomationapplocalbrokerhost
+- stage: 'msalautomationapplocalbrokerhost'
+  dependsOn: []    # this removes the implicit dependency on previous stage and causes this to run in parallel
+  displayName: Build MSAL Automation APKs (LocalBrokerHost)
```
somalaya, line 111: Fadi, what is the difference between msalautomationapp and msalautomationapplocalbrokerhost?  I did not understand what LocalBrokerHost means here.

somalaya (approved):

azure-pipelines/ui-automation/broker-test.yml
fadidurah: I'm creating another msal automation app with Local BrokerHost Variant. This is needed for some test cases to pass without needing to create another pipeline.

somalaya: Ok.. so this just means that msal automation app would run with broker as brokerHost app right?
I think we should mention that in one of the fields as "MSAL Automation app with broker set as local brokerHost app" and we should also update the PR description to indicate this change

fadidurah: Yeah, these changes came in after I initially made the Pr. Let me make some changes

fadidurah: The display name does include a detail mentioning BrokerHost

PR #172, opened 2023-01-04
Add script to promote packages in Ado maven feed

iamgusain:
### What
Adding a script that can automatically promote packages to different views in ADO maven feed.

### Why
CP team currently require us to promote our libs/pacakges to Prerelease/Release view in our internal Ado Maven feed before they can pull them in their feed. This is a manual process currently where an engineer has to manually promote all packages (total 6) to right views before they can be consumed by CP team. 
This is a blocker for our daily end to end pipeline which needs to generate a CP apk with our daily builds and run automation with it without requiring any manual step

### How
Added a new poweshell script that uses AzureDevOps REST API to promote packages.
We will consume this script in our daily pipeline to promote packages to "Prerelease" view

Another benefit is that the script can also be run standalone to promote packages, so it can be used to promote private builds in one step, as compared to promoting multiple packages individually from the UI

### Testing
verified locally by running the script to promote packages.

### DIFF
--- a/azure-pipelines/scripts/promote-packages.ps1
+++ b/azure-pipelines/scripts/promote-packages.ps1
@@ -1,0 +1,48 @@
+Param (
+    [Parameter(Mandatory = $true)][String]$PackagingPAT,
+    [Parameter(Mandatory = $false)][String]$PackageVersion = "",
+    [Parameter(Mandatory = $false)][String]$common4jVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$commonVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$broker4jVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$adAccountsVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$msalVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$adalVersion = $PackageVersion,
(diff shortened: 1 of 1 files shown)


azure-pipelines/scripts/promote-packages.ps1
```diff
@@ -0,0 +1,48 @@
+Param (
+    [Parameter(Mandatory = $true)][String]$PackagingPAT,
+    [Parameter(Mandatory = $false)][String]$PackageVersion = "",
+    [Parameter(Mandatory = $false)][String]$common4jVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$commonVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$broker4jVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$adAccountsVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$msalVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$adalVersion = $PackageVersion,
+    [Parameter(Mandatory = $false)][String]$PromoteToView = "Prerelease",
```
shahzaibj, line 10: Do they need it to be `release` instead or is `Prerelease` good enough?

shahzaibj (approved):
This is great. Thanks @iamgusain

azure-pipelines/scripts/promote-packages.ps1
iamgusain: I have requested them to use Prerelease.

PR #209, opened 2023-05-16
Fail Fast doesn't work base gcloud firebase command, only flank

fadidurah:
Can't use Fail-fast without flank, at least not that i could find. Might consider moving all ui automation execution to flank and use 1 shard for tests that don't actually need sharding.


rpdome:
can we revert the original PR instead?

fadidurah:
Reverting the other PR would also remove all the other changes made in that PR, which we actually want. I just need to remove --fail-fast from non-flank gcloud command calls

rpdome (approved):

PR #210, opened 2023-05-19
Build and install MsalTestApp on firebase device

cikelengfeng:
In order to enable E2E testing to verify the latest modifications of the msal sdk and broker sdk, we have added support for building and installing msalTestApp in the E2E testing process.

- Changed the “auth-client-android-dev” pipeline to pass the version of Msal SDK when trigger the “Android Broker CI” pipeline. 
- Changed the “Android Broker CI” pipeline to be able to trigger the maslTestApp pipeline with the specific msal sdk version.
- Install msalTestApp on firebase devices:
  - msal_with_broker_high_api in the broker-test.yml file 
  - msal_with_brokerhost_high_api in the broker-test.yml file 
  - msal_with_broker_low_api in the broker-test.yml file

### DIFF
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -177,7 +177,6 @@ stages:
         - publish: $(Build.ArtifactStagingDirectory)/msaltestapp
           displayName: 'Publish Msal Test apk for later use'
           artifact: msalTestApp
-          overwrite: true
         - task: UniversalPackages@0
           displayName: 'Publish Msal Test apk to internal feed'
           inputs:
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -49,6 +49,7 @@ variables:
   OSServiceConnection: LTW-Integration
   LTWFeedName: Auth-Broker-Integrated-LTW-Build
   LTWApk: LTW-signed.apk
+  msalSdkVersion: 0.0.20230517-dev.8
   
 parameters:
 - name: firebaseDeviceIdHigh
@@ -149,14 +150,14 @@ stages:
         - checkout: self
           persistCredentials: True
         - script: echo "The current user is $(Build.RequestedFor) $(Build.RequestedForId)"
-        - script: echo "The current msal version is ${{ variables['msalSdkVersion'] }}"
+        - script: echo "The current msal version is $(variables.msalSdkVersion)"
         - task: PowerShell@2
           displayName: Queue and wait for Msal Test App Apk generation pipeline
           name: buildApk
           continueOnError: true
           inputs:
             filePath: '$(Build.SourcesDirectory)/azure-pipelines/scripts/queue-build.ps1'
-            arguments: '-OrganizationUrl "https://identitydivision.visualstudio.com/" -Project "Engineering" -PipelinePAT "$env:SYSTEM_ACCESSTOKEN" -WaitTimeoutInMinutes 120 -BuildDefinitionId "1338" -PipelineVariablesJson "{ ''msal_version'': ''${parameters.msalSdkVersion}'' }" -Branch "${Build.SourceBranchName}"'
+            arguments: '-OrganizationUrl "https://identitydivision.visualstudio.com/" -Project "Engineering" -PipelinePAT "$env:SYSTEM_ACCESSTOKEN" -WaitTimeoutInMinutes 120 -BuildDefinitionId "1338" -PipelineVariablesJson "{ ''msal_version'': ''$(variables.msalSdkVersion)'' }" -Branch "${Build.SourceBranchName}"'
             workingDirectory: '$(Build.SourcesDirectory)'
           env:
             SYSTEM_ACCESSTOKEN: $(System.AccessToken)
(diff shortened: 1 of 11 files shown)


azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -29,10 +30,11 @@ variables:
   engineeringProjectId: 'fac9d424-53d2-45c0-91b5-ef6ba7a6bf26'
   azureSamplePipelineId: 1458
   brokerHostPipelineId: 1432
+  msalTestAppPipelineId: 1338
   msazureServiceConnection: AndroidBroker-CI
   msazureFeedName: Android-Broker
   msalApp: msalautomationapp-dist-AutoBroker-debug.apk
-  msalTestApp: msalautomationapp-dist-AutoBroker-debug-androidTest.apk
+  msalTestApp: msalTestApp-local-debug.apk
```
somalaya, line 37: I am not sure if we should be replacing this variable. @fadidurah to confirm if this change can be made?

azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -137,7 +139,54 @@ stages:
         brokerApp: AutoBroker
         brokerFlavor: Dist
         brokerSource: LocalApk
-# Download First Party Apps
+# TestAppApk - MsalTestApp Queue pipeline
+- stage: 'msalTestApp'
+  displayName: Build Msal Test App Apks
```
somalaya, line 144: Are we building multiple MSAL test apks? Or is it just the latest one MSALTestApp built in this stage? Let's rename this display name accordingly

azure-pipelines/ui-automation/broker-test.yml
fadidurah: No this variable is needed for other things. I don't think it will work if you change this name. Probably should just make a new variable

azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -137,7 +139,54 @@ stages:
...
+  dependsOn: []   # this removes the implicit dependency on previous stage and causes this to run in parallel
+  jobs:
+    - job: queue_build_MsalTestApp
```
fadidurah, line 147: This is a little odd to me. We are already building everything else in daily pipeline, why not also build the msal test app in the daily pipeline, and just fetch it in Broker CI pipeline like we do all the other apks? That way you wouldn't even need to pass msal version variable to Broker CI Pipeline.

somalaya: Dean, we just need to do the same thing we are already doing for building LTW apk in daily pipeline and then downloding the apk in the last stage "E2E UI automation" (Android broker CI pipeline)

cikelengfeng: MsalTestApp is built for e2e test, I think we should't build it in daily pipeline, because we don't always run e2e test in daily pipeline.
I also see we build MsalAutomationApp in Broker CI pipeline, I think building MsalTestApp should be in the same place with building MsalAutomationApp.

azure-pipelines/ui-automation/broker-test.yml
cikelengfeng: To distinguish them, I add a new variable named "msalE2ETestApp"

azure-pipelines/ui-automation/broker-test.yml
cikelengfeng: changed to "Build Msal Test App Apk"

azure-pipelines/ui-automation/broker-test.yml
fadidurah: MsalTestApp has usages outside of UI automation though, someone could pull it from the pipeline and use it for manual testing for a specific MSAL version regardless of their local environment (same thing for adal test app), where as the automation app is only useful for UI automation and is irrelevant to everything else. I still think building MsalTestApp in daily pipeline is more suitable.

azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -216,7 +269,8 @@ stages:
                       /sdcard/BrokerHost.apk=/home/vsts/work/1/brokerapks/$(brokerhost_apk),\
                       /data/local/tmp/test/DirectPushBrokerHost.apk=/home/vsts/work/1/brokerapks/$(brokerhost_apk),\
                       /data/local/tmp/DirectPushBrokerHost.apk=/home/vsts/work/1/brokerapks/$(brokerhost_apk),\
-                      /sdcard/Edge.apk=$(Pipeline.WorkSpace)/firstpartyapks/$(edgeApk)"
+                      /sdcard/Edge.apk=$(Pipeline.WorkSpace)/firstpartyapks/$(edgeApk),\
+                      /sdcard/MsalE2ETestApp.apk=$(Pipeline.WorkSpace)/msalE2ETestApp/$(msalE2ETestApp)"
```
fadidurah, line 273: Is this `MsalE2ETestApp.apk` the name that we are using in the UI Automation library? I feel this may cause some confusion, maybe simply "MSALTestApp" is simpler.

azure-pipelines/ui-automation/broker-test.yml
cikelengfeng: Maybe it is better to trigger the [MsalTestApp pipeline](https://identitydivision.visualstudio.com/Engineering/_build?definitionId=1338&_a=summary) directly in this scenario, I feel running the whole daily pipeline to build an MsalTestApp apk seems too heavy.

fadidurah (approved):

somalaya (approved):

PR #215, opened 2023-06-01
Running instrumented tests as part of daily pipeline

iamgusain:
Adding a job to run instrumented tests as part of daily pipeline run.
We have good set of tests that run as part of instrumented tests, making them part of daily pipeline run, so that we can pay close attention to when something starts failing in these tests.

The new job will run instrumented tests for Common. Broker, Adal and Msal
Here is a sample run: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1108674&view=results

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -67,14 +67,14 @@ project(':uiautomationutilities').projectDir = new File('common/uiautomationutil
 include(":brokerautomationapp")
 project(':brokerautomationapp').projectDir = new File('broker/brokerautomationapp')
 
-include(":oneauth")
-project(':oneauth').projectDir = new File('oneauth/sources/android/gradle')
-
-include(":oneauthtestapp")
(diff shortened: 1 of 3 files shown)


settings.gradle
```diff
@@ -67,14 +67,14 @@ project(':uiautomationutilities').projectDir = new File('common/uiautomationutil
 include(":brokerautomationapp")
 project(':brokerautomationapp').projectDir = new File('broker/brokerautomationapp')
 
-include(":oneauth")
```
fadidurah, line 70: Intended?

fadidurah (approved):

settings.gradle
iamgusain: Yes, since we don't use these regularly as part of android-complete, keeping them commented by default

fadidurah: Works for me

PR #216, opened 2023-06-01
Updating pipelines to use appropriate library versions when running UI Automation

fadidurah:
Recently we made some changes to brokerautomationapp that changed the dist library version used for adal to use the most recent released version, rather than the daily pipeline version. This PR includes changes to utilize the -Pdist command line parameters in brokerautomationapp and msalautomationapp to pass the correct library versions depending on which pipeline is running automation.

For daily pipeline, we use 0.0.+
for release pipeline, we use the lib versions passed as pipeline parameters.

Keep in mind that these versions only encompass msal and adal version in their respective automation app. broker version is not handled in this way, as it is passed directly when building the broker apps used in our tests

Passing pipelines
https://dev.azure.com/IdentityDivision/Engineering/_build/results?buildId=1109356&view=results
https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1109359&view=results


iamgusain (approved):
:shipit:

PR #228, opened 2023-06-29
Pass common version to one auth

somalaya:
To build oneAUth test app, we are triggering OneAuthTestApp pipeline from the Android CI pipeline. But previously, we were not passing the latest common version to be used while building OneAuthTestApp. With this change, we pass the common version to OneAuthTestApp pipeline.

### DIFF
--- a/azure-pipelines/scripts/queue-build.ps1
+++ b/azure-pipelines/scripts/queue-build.ps1
@@ -4,6 +4,7 @@ Param (
     [Parameter(Mandatory = $true)][String]$PipelinePAT,
     [Parameter(Mandatory = $true)][String]$BuildDefinitionId,
     [Parameter(Mandatory = $false)][String]$PipelineVariablesJson,
+    [Parameter(Mandatory = $false)][String]$TemplateParams,
     [Parameter(Mandatory = $false)][String]$Branch,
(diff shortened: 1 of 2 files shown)


azure-pipelines/scripts/queue-build.ps1
```diff
@@ -4,6 +4,7 @@ Param (
     [Parameter(Mandatory = $true)][String]$PipelinePAT,
     [Parameter(Mandatory = $true)][String]$BuildDefinitionId,
     [Parameter(Mandatory = $false)][String]$PipelineVariablesJson,
+    [Parameter(Mandatory = $false)][String]$TemplateParams,
```
fadidurah, line 7: Could you have used PipelineVariablesJson instead of this new parameter?

fadidurah (approved):

azure-pipelines/scripts/queue-build.ps1
somalaya: No, I cannot use PipelineVariablesJson as it is also being populated into parameters object. And I only want to populate this to templateParameters

fadidurah: sounds good

PR #230, opened 2023-06-30
add spot bugs template

p3dr0rv:


fadidurah (approved):
Do we use this template anywhere yet?

p3dr0rv:
> Do we use this template anywhere yet?

we use it for all PR pipelines, but the template is in common, So I want t to move the template to Android-complete, because it is easy to make modifications on this repo

PR #231, opened 2023-07-03
Build MsalTestApp with Dist flavor and Release signing configuration

cikelengfeng:
The MsalTestApp pipeline only consume msal_sdk_version in Dist flavor, and the Msal SDK passed from "Broker CI" pipeline is built in Release mode, so we should trigger the MsalTestApp pipeline with not only msal_sdk_version but also Dist flavor and Release signing configuration.

### DIFF
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -183,7 +183,7 @@ stages:
           continueOnError: true
           inputs:
             filePath: '$(Build.SourcesDirectory)/azure-pipelines/scripts/queue-build.ps1'
-            arguments: '-OrganizationUrl "https://identitydivision.visualstudio.com/" -Project "Engineering" -PipelinePAT "$env:SYSTEM_ACCESSTOKEN" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(msalE2ETestAppPipelineId)" -PipelineVariablesJson "{ ''msal_version'': ''$(msal_sdk_version)'' }" -TemplateParams "{''signingConfigurations'': ''Release''}" -Branch "$(Build.SourceBranch)"'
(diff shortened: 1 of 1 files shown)


azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -183,7 +183,7 @@ stages:
           continueOnError: true
           inputs:
             filePath: '$(Build.SourcesDirectory)/azure-pipelines/scripts/queue-build.ps1'
-            arguments: '-OrganizationUrl "https://identitydivision.visualstudio.com/" -Project "Engineering" -PipelinePAT "$env:SYSTEM_ACCESSTOKEN" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(msalE2ETestAppPipelineId)" -PipelineVariablesJson "{ ''msal_version'': ''$(msal_sdk_version)'' }" -Branch "$(Build.SourceBranch)"'
+            arguments: '-OrganizationUrl "https://identitydivision.visualstudio.com/" -Project "Engineering" -PipelinePAT "$env:SYSTEM_ACCESSTOKEN" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(msalE2ETestAppPipelineId)" -PipelineVariablesJson "{ ''msal_version'': ''$(msal_sdk_version)'' }" -TemplateParams "{''signingConfigurations'': ''Release'', ''productFlavors'': ''Dist''}" -Branch "$(Build.SourceBranch)"'
```
somalaya, line 186: is there any reason why we are using templateParams instead of  pipelineVariablesJson which would translate to parameters?

cikelengfeng: Because "signingConfigurations" and "productFlavors" are pipeline parameters defined in MsalTestApp pipeline, passing these arguments to pipelineVariablesJson won't work.

somalaya (approved):

PR #243, opened 2023-08-15
Target, compile SDK, AGP and gradle version increase

somalaya:
1. Target sdk version changed to 33
2. Compile sdk version changed to 34
3. AGP version changed to 7.4.2
4. Gradle distribution url changed to 7.5

After changing above versions, I had to update other dependencies for compatibility

1. Powerlift version changed to 1.0.3
2. Bouncy castle version changed to 1.69
3. robolectric version to 4.9
4. Also, with new gradle versions, some of the properties used in gradle were deprecated (naming conventions changed). So, made those changes
5. With targetSDK changes, we should not be having a package name in AndroidManifest.xml.. instead, should add a namespace property in build.gradle. https://developer.android.com/about/versions/14/behavior-changes-14#safer-intents
6. Test apps only build with Java 11 when gradle version is 7.5. So I set a jdkVersionOption  parameter in gradle https://developer.android.com/build/releases/past-releases/agp-7-0-0-release-notes

MicrosoftTeams-image (8)

7. Removed SonarQube Quality gate analysis because it was not being used - Was set to default. Also, we have CodeQL that is doing the Quality gate analysis.
8. Update gradle dependency lockfile path https://docs.gradle.org/7.0/userguide/dependency_locking.html#generating_and_updating_dependency_locks

How did I validate if these changes break any of the existing scenarios?
1. I triggered the daily pipeline to verify if all the UTs and instrumented tests are succeeding 
2. Also ran UI tests with auth app and CP generated from consuming the common, msal and broker from our daily pipeline. Those tests are also succeeding


azure-pipelines/continuous-delivery/assemble&publish.yml
```diff
@@ -75,15 +75,15 @@ jobs:
       ${{ parameters.vstsMvnAndroidAccessToken }}: $(System.AccessToken)
   - task: ComponentGovernanceComponentDetection@0
     inputs:
-      sourceScanPath: $(Build.SourcesDirectory)/${{ parameters.project }}/gradle/dependency-locks/
+      sourceScanPath: $(Build.SourcesDirectory)/${{ parameters.project }}/
```
iamgusain, line 78: is the path changed for the lockfile?

azure-pipelines/test-app/msal-test-app.yml
```diff
@@ -37,18 +37,18 @@ resources:
   - repository: common
     type: github
     name: AzureAD/microsoft-authentication-library-common-for-android
-    ref: dev
+    ref: somalaya/testAGP
```
iamgusain, line 40: revert before merging

azure-pipelines/ui-automation/templates/build-msal-test-app.yml
```diff
@@ -24,6 +24,7 @@ parameters:
       - PROD
       - RC
 
+
```
iamgusain, line 27: nit: remove

azure-pipelines/continuous-delivery/assemble&publish.yml
somalaya: yes, I do not see it being generated in gradle/dependencies anymore.

azure-pipelines/test-app/msal-test-app.yml
somalaya: Sure, I kept this for testing and running the pipelines.  Let me leave this comment open until all the review is completed.

somalaya: reverted

azure-pipelines/continuous-delivery/assemble&publish.yml
iamgusain: discussed this offline, we can publish the individual lockfile (gradle.lockfile) from the new location, instead of publishing the full folder.

iamgusain (approved):

azure-pipelines/continuous-delivery/assemble&publish.yml
somalaya: I will do this as a separate task.

PR #254, opened 2023-09-20
Fix LTW generation in pipeline

cikelengfeng:
LTW building pipeline need a special build reason "UserCreated" to indicate that a pipeline run is triggered by Broker pipeline, then LTW pipeline can do some additional jobs for Broker pipeline. (e.g. push LTW apk to specific azure feed so that Broker E2E automation tests can download it).

However, the build reason was changed from "UserCreated" to "ResourceTrigger" in this [PR](https://github.com/AzureAD/android-complete/pull/250). 

My PR make this build reason to be a parameter so that daily pipeline can specify it to "UserCreated" for LTW generation and keep "ResourceTrigger" as default value.


cikelengfeng:
> why does LTW pipeline need the BuildReason to be UserCreated? This means that the build is triggered by the user and daily pipeline is not necessarily triggered by any user (as it has scheduled runs)

LTW pipeline need this build reason to indicate which run is triggered by Broker pipeline (this might be tricky, but I need to keep it for now to save time and minimize change)

somalaya (approved):

PR #255, opened 2023-09-22
Create another schedule pipeline for Flighted library validation

fadidurah:
**What**
Adding infra to support running our daily pipeline with flights. Supporting both flighted and non-flighted runs in one pipeline would make things pretty messy, as many steps would have to be duplicated for flighted vs non-flighted. Instead, I created a new Pipeline, that will run every Monday, Wednesday, Friday at 4AM.
https://identitydivision.visualstudio.com/Engineering/_build?definitionId=2044&_a=summary

Also took time to update some PAT references. We currently use an MSAzure PAT for queuing CP and AuthApp builds, as well as a PAT for queuing LTW pipeline. I've since added these to our keyvault in azure portal and have added them to the AndroidAuthClientAutomationSecrets variable group. This is a more secure approach, and any pipelines that need to access these tokens can use the variable group rather than declaring them again in the pipeline's variable page.

Related Broker PR to enable build parameters for local flights https://github.com/AzureAD/ad-accounts-for-android/pull/2548

**Why**
This will enable us to validate our libraries against our flights, as well as enabling future flighting validation and unblocking LTW testing effort.

**How**
Update the yml script to use the correct PATs and add support for passing flights. Also added a new YML for the second pipeline, as well as a new assemble and publish template to support flighting. This is temporary, as we cannot use the system token in this new pipeline because of an unknown issue. This will be addressed later, and both pipelines will use the same yml in the future once we've solved this issue, to make it easier to update the script. PBI to revisit this in November https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2701835

**Testing**
Base Daily Pipeline should still work as intended. Test run https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1180381&view=results
image

New pipeline should work as well, with properly flighted APKs. https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1180380&view=results
image


azure-pipelines/continuous-delivery/local-flight-assemble&publish.yml
fadidurah, line 1: Ignore this one, it was not being used and was an older implementation i had merged.

azure-pipelines/continuous-delivery/flighted-assemble&publish.yml
```diff
@@ -0,0 +1,161 @@
...
+          dos2unix gradlew
+          chmod +x gradlew
+          export DISPLAY=:0.0
+          eval $(dbus-launch --sh-syntax)
+          sudo apt install gnome-keyring
+          /usr/bin/gnome-keyring-daemon --start --components=secrets
+          ./gradlew LinuxBroker:linuxBrokerUnitTestCoverageReport ${{ parameters.testParams}}
+      env:
+        ${{ parameters.vstsMvnAndroidUsername }}: $(AndroidAuthClient-Internal-Maven-Username)
+        ${{ parameters.vstsMvnAndroidAccessToken }}: $(FadiPAT)
```
fadidurah, line 138: Had a long sync with Amit to try and get the system token working in this pipeline, but we could not figure out why we were unable to push to maven with that token. Have to use a PAT i created for this instead.

azure-pipelines/continuous-delivery/flighted-auth-client-android-dev.yml
p3dr0rv, line 1: do we need this file?, this is just a copy of auth-client-android-dev.yml
I personally don't like the idea of having to instances of this humongous file. if we need to update something we will need to change 2 files

azure-pipelines/ui-automation/templates/build-broker-host.yml
```diff
@@ -36,6 +36,12 @@ parameters:
   values:
     - PROD
     - RC
+- name: localFlightsParamString
+  type: string
+  default: ''
+- name: flagsParamString
```
p3dr0rv, line 42: what is this variable for?

p3dr0rv:
I suggest that instead of creating two files one for PAT and one for System.AT, use a flag to determine what AT to use and make System.AT as default.

fadidurah:
> I suggest that instead of creating two files one for PAT and one for System.AT, use a flag to determine what AT to use and make System.AT as default.

I considered this, but i did not want to pass Access Tokens across templates as that may end up exposing them through pipeline logs, and if i were to create conditional statement throughout the yml file to determine which token variable to use, it would end up being a lot of throw away work. I think this is the quickest way to do this for now, and will not leave any clean up work for when i eventually use the same yml for both files.

azure-pipelines/ui-automation/templates/build-broker-host.yml
fadidurah: We can pass flags during brokerhost build, i.e. -PenablePrtv3

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
```diff
@@ -355,10 +359,12 @@ stages:
           displayName: Queue and wait for Authenticator Apk generation pipeline
           inputs:
             filePath: '$(Build.SourcesDirectory)/azure-pipelines/scripts/queue-build.ps1'
-            arguments: '-OrganizationUrl "https://msazure.visualstudio.com/" -Project "One" -PipelinePAT "$(pipelinepat)" -BuildDefinitionId "237369" -PipelineVariablesJson "{ ''AdAccountsVersion'': ''$(brokerVersionNumber)'', ''CommonVersion'': ''$(versionNumber)'', ''MsalVersion'': ''$(versionNumber)'', ''AdalVersion'': ''$(versionNumber)'' }" -Branch "$(authenticatorBranch)"'
+            arguments: '-OrganizationUrl "https://msazure.visualstudio.com/" -Project "One" -PipelinePAT "$(MSAzure-PAT)" -BuildDefinitionId "237369" -PipelineVariablesJson "{ ''AdAccountsVersion'': ''$(brokerVersionNumber)'', ''CommonVersion'': ''$(versionNumber)'', ''MsalVersion'': ''$(versionNumber)'', ''AdalVersion'': ''$(versionNumber)'' }" -Branch "$(authenticatorBranch)"'
```
iamgusain, line 362: Can you also define the buildDefinitionId as variable at the top, and add a comment which pipeline it is pointing to.
# Variable: 'authenticatorPipelineId' pipeline id for Authenticator-Pipeline to generate Authenticator APK with build libraries https://msazure.visualstudio.com/One/_build?definitionId=237369&_a=summary

iamgusain: This will make it easier for future references.

azure-pipelines/continuous-delivery/flighted-assemble&publish.yml
iamgusain: let's add a workitem to investigate and fix this.

azure-pipelines/ui-automation/templates/build-broker-host.yml
iamgusain: can we add all the flags, flights that are passed as tags to the pipeline, so we can easily see which flags the pipeline ran with.

iamgusain: something like this.
image

azure-pipelines/continuous-delivery/flighted-assemble&publish.yml
```diff
@@ -0,0 +1,161 @@
+# File: azure-pipelines/continuous-delivery/assemble&publish.yml
```
iamgusain, line 1: is there a way we can reuse assemble&publish.yml, or create a template with shared code.
So that we don't have to make changes in future in two places, as it is very easy to only change in one place and miss updating in another.

azure-pipelines/continuous-delivery/flighted-assemble&publish.yml
fadidurah: Yup https://identitydivision.visualstudio.com/Engineering/_workitems/edit/2701835

azure-pipelines/continuous-delivery/flighted-auth-client-android-dev.yml
fadidurah: Since this comment, we've had to make some accommodations to make using the second pipeline easier for LTW folks, some parameters have different default values between the two parameters. So we're going to need a different yml script anyway for now. I can merge them both once LTW testing effort is complete.

fadidurah:
> I suggest that instead of creating two files one for PAT and one for System.AT, use a flag to determine what AT to use and make System.AT as default.

I can implement this for the assemble and publish file

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
fadidurah: Sure

p3dr0rv:
> across templates as that may end up exposing them through pipeline logs. There are also some differences between the files that are needed by LTW folks for now.

you do not need to expose the token System.AT can be accessed from any template and for PAT you can set this token in the environment, so there is no need to log it or pass it.

What other differences? the time I review the only diff was the token

azure-pipelines/ui-automation/templates/build-broker-host.yml
fadidurah: Done

fadidurah:
> > across templates as that may end up exposing them through pipeline logs. There are also some differences between the files that are needed by LTW folks for now.
> 
> you do not need to expose the token System.AT can be accessed from any template and for PAT you can set this token in the environment, so there is no need to log it or pass it.
> 
> What other differences? the time I review the only diff was the token

Default values for some of the boolean parameters of the pipeline (Run UI Validation and the broker selection boolean)

fadidurah:
Removed references to FadiPAT, now called mvnPAT

fadidurah:
> > across templates as that may end up exposing them through pipeline logs. There are also some differences between the files that are needed by LTW folks for now.
> 
> you do not need to expose the token System.AT can be accessed from any template and for PAT you can set this token in the environment, so there is no need to log it or pass it.
> 
> What other differences? the time I review the only diff was the token

Addressed in teams sync

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
```diff
@@ -82,19 +83,22 @@ parameters:
...
 
 variables:
-  ${{ if eq(parameters.customVersionNumber, 'Default') }}:  
+  ${{ if eq(parameters.customVersionNumber, 'Default') }}:
     versionNumber: $(Build.BuildNumber)
     brokerVersionNumber: $[replace(variables['build.buildnumber'], '1.0.', '8.1.')]
   ${{ else }}:
     versionNumber:  ${{ parameters.customVersionNumber }}
-    brokerVersionNumber: ${{ parameters.customVersionNumber }}
+    brokerVersionNumber: $[replace(${{ parameters.customVersionNumber }}, '1.0.', '8.1.')]
```
iamgusain, line 91: we don't need it in this case, the custom version number is same for all libs.

iamgusain (approved):

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
fadidurah: Yes, I'm thinking in terms to using the pipeline to generate broker APKs from existing library versions, say we had a run with `1.0.20230928-dev.9`. The CP and authenticator builds would always fail, because they would try to pull android broker version 1.0.20230928-dev.9, instead of 8.1.20230928-dev.9.

This way they would pull the correct version. Thoughts?

fadidurah: This shouldn't affect any other use case, unless a user is trying to build versions of our libs that include "1.0." specifically. Which i think is a very small use case.

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
fadidurah: Addressed https://github.com/AzureAD/android-complete/pull/255/commits/1352146723ccfd3b4f2c11764a455b2fac1315bb

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
iamgusain: I see so your use case is if we use the pipeline to just generate the broker APKs and skip the build and assemble.
Not sure if we are using daily pipeline for that purpose. But I am fine with this change.

docker/api29/Dockerfile
fadidurah, line 1: Deleted this and the other file to avoid this error: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1182948&view=logs&j=d21cf4c0-4b13-5059-6e01-f7633376be4f&t=30a4fd42-5c09-5f49-7a07-4cb254e9a89e

Success after deletion https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1182967&view=results

PR #266, opened 2023-10-26
Change agent in RunUnitTest job

p3dr0rv:
WHY?

running test with ubuntu agents usually ends with OOM exceptions.
this change allows to use custom agents.

### DIFF
--- a/azure-pipelines/continuous-delivery/assemble&publish.yml
+++ b/azure-pipelines/continuous-delivery/assemble&publish.yml
@@ -104,7 +104,10 @@ jobs:
   timeoutInMinutes: 120
   continueOnError: true
   pool:
-    vmImage: ubuntu-latest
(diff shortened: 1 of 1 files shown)


azure-pipelines/continuous-delivery/assemble&publish.yml
```diff
@@ -104,7 +104,10 @@ jobs:
   timeoutInMinutes: 120
   continueOnError: true
   pool:
-    vmImage: ubuntu-latest
+    ${{ if eq(parameters.agentImage, 'none') }}:
+      vmImage: ubuntu-latest
+    ${{ else }}:
+      name: ${{ parameters.agentImage }}
```
somalaya, line 110: do we need to pass custom agents for our UTs from now on in daily pipeline?

PR #267, opened 2023-10-26
Fix push to feed in flighted pipeline, and enable wildcard to work when pulling from feed

fadidurah:
For unknown reasons, flighted pipeline seems to not have permission to publish to artifact feed, but original daily pipeline does. This PR includes a change to use the PAT in universal package upload. Also remove coverage report combination step, this never worked to begin with.
Also updated how brokerhost versioning was handled to enable using wildcard in broker CI pipeline


azure-pipelines/continuous-delivery/flighted-auth-client-android-dev.yml
```diff
@@ -555,7 +531,7 @@ stages:
...
+            arguments: '-OrganizationUrl "https://identitydivision.visualstudio.com/" -Project "Engineering" -PipelinePAT "$env:SYSTEM_ACCESSTOKEN" -BuildDefinitionId $(brokerAutomationPipelineId) -BuildIdOutputVar "brokerAutomationBuildId" -Branch  "$(Build.SourceBranch)" -PipelineVariablesJson "{ ''dailyVersion'' : ''$(versionNumber)'', ''test_run_prefix'' : ''(Flighted, LTW) '', ''msalTestTarget'' : ''${{ parameters.msalTestTarget }}, annotation com.microsoft.identity.client.ui.automation.annotations.LTWTests'', ''brokerTestTarget'' : ''${{ parameters.brokerTestTarget }}, annotation com.microsoft.identity.client.ui.automation.annotations.LTWTests'', ''msal_branch'' : ''$(MSAL_LTW_Branch)'', ''broker_branch'' : ''$(Broker_LTW_Branch)'', ''msal_sdk_version'': ''$(versionNumber)'', ''commonVersion'': ''$(versionNumber)'', ''BrokerHostFeedPackageName'': ''com.microsoft.identity.testuserapp.flighted''}"'
             workingDirectory: '$(Build.SourcesDirectory)'
```
somalaya, line 535: could you help understand why we have com.microsoft.identity.testuserapp.flighted as opposed to com.microsoft.identity.testuserapp?

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
```diff
@@ -106,6 +106,7 @@ variables:
   ${{ else }}:
     enableBrokerSelectionParam: ''
     enableBrokerDiscoveryParam: ''
+  brokerHostPackageVersionCounter: $[counter(1, 10000)]
```
somalaya, line 109: So the brokerHost version format would be 1.0.1 to 1.0.10000?

azure-pipelines/continuous-delivery/flighted-auth-client-android-dev.yml
fadidurah: I decided to push the flighted and non-flighted brokerhosts to distinct packages in the feed, to create more separation

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
fadidurah: Yes, but the counter itself only accounts for the 10000 part

somalaya: cool, so what happens after 10000 publishes? would it start from 1 again? I believe we don't delete the old apks right?

azure-pipelines/continuous-delivery/flighted-auth-client-android-dev.yml
somalaya: why was this distinction required? Is it useful while pulling/downloading the apks in Android CI pipeline?

azure-pipelines/continuous-delivery/auth-client-android-dev.yml
fadidurah: it will simply keep incrementing, so it would be 20000. It shouldn't ever loop back

azure-pipelines/continuous-delivery/flighted-auth-client-android-dev.yml
fadidurah: I find it useful, this way we make sure we are always getting flighted version. if you push to the same feed, and try to pull the latest, you might end up with a non-flighted version when you dont want it, and vise versa. This makes things easier imo

somalaya (approved):

PR #272, opened 2023-11-14
Improve APK pulling from Universal Packages in Daily Pipelines

fadidurah:
We currently use wildcard "\*" version numbers to pull APKs from universal packages, and give the pipeline runner the ability to use a specific version if desired. "\*" matches to the APK with the highest version number. The assumption was that if daily pipeline generates an authenticator or cp build, and then calls UI automation pipeline right away, we are going to get those generated builds because they were just uploaded to the feed and therefore should have the highest version number. This proved to be a faulty assumption, for two reasons. The first, between the APK generation and the call to the UI validation pipeline in our daily runs, it's possible for new APKs to have been pushed to the feed, especially as more people are using the pipelines for various reasons. The second, the most recently uploaded apk is not always the one with the highest version, so * would not pick it up. The changes in this PR address this issue.

In this PR, i make changes to enable our pipelines to fetch the build number from the broker APK queue pipelines (Authenticator, Company Portal, LTW). The build numbers from these pipelines correspond to the apk version numbers used when publishing to universal packages, allowing us to use these versions as variables in future stages/ other pipelines when pulling from universal packages, to ensure that we are always pulling the correct versions from the feed. If the daily pipeline doesn't generate a particular apk (say we ran without generating CP), we won't pass that apks version to the CI pipeline, and we fallback on the wildcard "\*".

Additional changes:
- We also pull the broker apks in our daily runs, so that they are accessible in the artifacts tab.
- Add boolean to disable publishing release and lockfile artifacts, reduces clutter in pipelines.

Validation:

Here's a daily run with these changes, you'll notice new tags being set with the apk versions generated in the pipeline. These are then passed to the CI pipeline as parameters.
image

This is a screenshot of the CI pipeline following the daily build, showing that the specific versions generated are sent as parameters to the CI pipeline. (Pipeline link https://dev.azure.com/IdentityDivision/Engineering/_build/results?buildId=1209711&view=results)
image

Daily pipeline test: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1209557&view=results
Flighted Pipeline test: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1209558&view=results

cc. @rpdome and @shahzaibj, as these changes address some asks proposed by them in the past


somalaya:
would the same version numbers be used when we retrigger the pipeline? Or do we need to manually fill the versions?

shahzaibj:
On the tags here or somewhere on the UX: https://user-images.githubusercontent.com/88730756/283203672-a057f2b2-e84f-4a50-95c4-1b9959de86c9.png

Can we also add/include links to build runs for AuthApp / CP / LTW? (So the engineer can easily navigate to those builds)

fadidurah:
> would the same version numbers be used when we retrigger the pipeline? Or do we need to manually fill the versions?

for the CI Pipeline? The parameters should persist between re-runs until the user changes them

fadidurah:
> On the tags here or somewhere on the UX: https://user-images.githubusercontent.com/88730756/283203672-a057f2b2-e84f-4a50-95c4-1b9959de86c9.png
> 
> Can we also add/include links to build runs for AuthApp / CP / LTW? (So the engineer can easily navigate to those builds)

Not sure if i can put a link in the tags, i can check to see if there's an easy way to point to the runs

rpdome (approved):

shahzaibj (approved):

PR #279, opened 2023-12-11
Add native auth dependency + dependency versions

SammyO:
- Add native auth sample app to `droidSetup` script
- Add native auth app to project settings
- Add native auth specific dependency versions to `versions.gradle`

note: native auth sample app is currently private. This means Github credentials are needed to download contents as part of `droidSetup`. Github doesn't support username + password, so this script uses SSH (which requires an SSH key). In February native auth will go to public preview and the respective sample app repo will become public. This `droidSetup` script can be updated then to pull from the HTTP source.

### DIFF
--- a/.gitconfig
+++ b/.gitconfig
@@ -22,6 +22,7 @@
 		git clone -b dev https://github.com/AzureAD/ad-accounts-for-android.git broker; \
 		git clone -b working https://msazure.visualstudio.com/DefaultCollection/One/_git/AD-MFA-phonefactor-phoneApp-android authenticator; \
         git clone -b master https://github.com/Azure-Samples/ms-identity-android-java.git azuresample; \
+        git clone -b main git@github.com:Azure-Samples/ms-identity-ciam-native-auth-android-sample.git nativeauthsample; \
         git clone -b dev https://office.visualstudio.com/DefaultCollection/OneAuth/_git/OneAuth oneauth; \
         git clone -b develop https://github.com/AzureAD/microsoft-authentication-library-for-cpp.git msalcpp; \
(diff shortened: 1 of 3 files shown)


gradle/versions.gradle
```diff
@@ -68,6 +69,16 @@ ext {
     yubikitAndroidVersion = "2.3.0"
     yubikitPivVersion = "2.3.0"
     powerliftAndroidVersion="1.0.3"
+    openTelemetryVersion = "1.18.0"
```
SammyO, line 72: I had to add these in order to get sample apps to compile, even though these are already defined in the project-specific `gradle/versions.gradle`. Not sure if Gradle looks only at the top level `versions.gradle` or also recursively into the projects' `versions.gradle`?

fadidurah (approved):

.gitconfig
```diff
@@ -22,6 +22,7 @@
 		git clone -b dev https://github.com/AzureAD/ad-accounts-for-android.git broker; \
 		git clone -b working https://msazure.visualstudio.com/DefaultCollection/One/_git/AD-MFA-phonefactor-phoneApp-android authenticator; \
         git clone -b master https://github.com/Azure-Samples/ms-identity-android-java.git azuresample; \
+        git clone -b main git@github.com:Azure-Samples/ms-identity-ciam-native-auth-android-sample.git nativeauthsample; \
```
GBakolas, line 25: the script uses https to clone the other repositories. Here we are using ssh is there a reason for this?
If we choose this way then the developer would need to configure a public ssh key to their github account for this to work

SammyO: See the note in the PR description. Does that answer your question?

GBakolas: Ok this explains everything. Sorry didn't see the note

GBakolas (approved):

PR #284, opened 2023-12-18
Fix step where oneauth test app is pulled from OneAuth pipeline

fadidurah:
A PAT recently expired in one of the existing oneauth connections. The connection was created by tanmay, but i don't have the permissions to update it. I've created a new connection, in which i've granted admin roles to multiple people in project and organization level, this makes it easier to edit in the future.

New connection: https://identitydivision.visualstudio.com/Engineering/_settings/adminservices?resourceId=cb1685ed-b09b-4a08-8ee0-44793edc38c3

Validation run: https://identitydivision.visualstudio.com/Engineering/_build/results?buildId=1225673&view=results

### DIFF
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -212,36 +212,36 @@ stages:
           displayName: 'Publish Msal Test apk for later use'
           artifact: msalE2ETestApp
 # TestAppApk - OneAuthTestApp Queue pipeline
-#- stage: 'oneAuthTestApp'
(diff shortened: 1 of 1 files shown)


iamgusain:
the previous one had admin access to everyone under AndroidAuth, did it not allow you to update.
image

fadidurah:
> the previous one had admin access to everyone under AndroidAuth, did it not allow you to update. image

I was not able to update this one, no. I think it's because permission is split between project and org, I guess i wasn't in org access

somalaya (approved):

PR #289, opened 2023-12-29
Update OneAuth integration branch to fix an issue with latest common

somalaya:
We recently made a change in common https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/2146 that broke OneAuth pipeline. Matthew has made some changes to fix it by updating their compileSDK version to 34 and a bunch of other changes in his branch. This PR simply points to his branch instead of an old integration branch which is not required anymore.

### DIFF
--- a/azure-pipelines/ui-automation/broker-test.yml
+++ b/azure-pipelines/ui-automation/broker-test.yml
@@ -228,7 +228,7 @@ stages:
           continueOnError: true
           inputs:
             filePath: '$(Build.SourcesDirectory)/azure-pipelines/scripts/queue-build.ps1'
-            arguments: '-OrganizationUrl "https://office.visualstudio.com/" -Project "OneAuth" -PipelinePAT "$(OfficePAT)" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(oneAuthTestAppPipelineId)" -Branch "android/common-ingestion" -TemplateParams "{''androidCommonVersion'': ''$(dailyVersion)''}"'
+            arguments: '-OrganizationUrl "https://office.visualstudio.com/" -Project "OneAuth" -PipelinePAT "$(OfficePAT)" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(oneAuthTestAppPipelineId)" -Branch "android/common-ingestion-with-compile-34" -TemplateParams "{''androidCommonVersion'': ''$(dailyVersion)''}"'
             workingDirectory: '$(Build.SourcesDirectory)'
         - task: DownloadExternalBuildArtifacts@15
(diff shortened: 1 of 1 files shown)


iamgusain:
Could this have been caught by our consumers of common pipeline?

azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -228,7 +228,7 @@ stages:
           continueOnError: true
           inputs:
             filePath: '$(Build.SourcesDirectory)/azure-pipelines/scripts/queue-build.ps1'
-            arguments: '-OrganizationUrl "https://office.visualstudio.com/" -Project "OneAuth" -PipelinePAT "$(OfficePAT)" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(oneAuthTestAppPipelineId)" -Branch "android/common-ingestion" -TemplateParams "{''androidCommonVersion'': ''$(dailyVersion)''}"'
+            arguments: '-OrganizationUrl "https://office.visualstudio.com/" -Project "OneAuth" -PipelinePAT "$(OfficePAT)" -WaitTimeoutInMinutes 120 -BuildDefinitionId "$(oneAuthTestAppPipelineId)" -Branch "android/common-ingestion-with-compile-34" -TemplateParams "{''androidCommonVersion'': ''$(dailyVersion)''}"'
```
iamgusain, line 231: Let's take the branch name as a variable/parameter with a default value, so it's easier to test and change.

somalaya:
> Could this have been caught by our consumers of common pipeline?

No, all the conumers of common pipeline already have compileSDK 34. OneAuth is currently in the process of updating the compileSDK to 34. That is why this issue happened.

iamgusain (approved):
Approved with suggestion.

azure-pipelines/ui-automation/broker-test.yml
somalaya: Sure, I will do this later as this needs to be merged to unblock others.

melissaahn:
> > Could this have been caught by our consumers of common pipeline?
> 
> No, all the conumers of common pipeline already have compileSDK 34. OneAuth is currently in the process of updating the compileSDK to 34. That is why this issue happened.

I thought this issue occurred because I set a newer version of androidx.lifecycle than what OneAuth was using? Which is why they were getting a deprecation warning on their side for a class that was considered deprecated in the newer version. I don't think the version of lifecycle that we're using depends on compileSdk 34. Correct me if i'm wrong here though.
If this is the case, is it still not possible to implement some sort of check in the consumers of common pipeline?

melissaahn (approved):

PR #292, opened 2024-01-05
Update old authenticator apk name

somalaya:
Updated old authenticator name to OldAuthenticator.apk and also uploaded a newer version of the apk


azure-pipelines/ui-automation/broker-test.yml
```diff
@@ -91,7 +91,7 @@ parameters:
 - name: oldAuthenticatorVersion
   displayName: Old Authenticator Version
   type: string
-  default: '6.2204.2470'
+  default: '6.2401.69'
```
iamgusain, line 94: This looks from this year, should we pick one from AuthApp Nov release

iamgusain (approved):

azure-pipelines/ui-automation/broker-test.yml
somalaya: I created this auth app build from a recent change made to broker selection logic. This is required for LTW tests.

PR #428, opened 2026-04-30
Add aria-alert-investigator skill for Aria health-metric IcMs, Fixes AB#3589793

somalaya:
Adds a new skill at `.github/skills/aria-alert-investigator/SKILL.md` that handles Aria-detected anomaly IcMs (titles starting with `Aria detected an incident in <project> for <metric>`).

## Why

These alerts have a fundamentally different investigation pattern than customer-reported incidents: there is no customer, no error chain, and no log file — only "the curve moved". The existing `incident-investigator` skill assumes customer-reported failures and produces wrong conclusions on Aria alerts. I have hit this several times now while triaging anomaly-detection IcMs.

## What the new skill enforces

- **Decode the metric → KQL slice (BLOCKING)** — Aria health-metric names are marketing-style labels; never run queries until the slice is confirmed (or assumption is explicitly flagged).
- **Treat the alert value as opaque** — don't reverse-engineer "4.48 in Red band" into "4 events".
- **Multi-angle trend analysis** — raw count, normalized rate, traffic volume, same-day-of-week, hourly-around-alert. A single view fools you.
- **Past ICMs are pattern signal, not verdict** — call out the family pattern, but don't conclude with it.
- **Code context is the baseline, not extra step** — delegate continuously to `codebase-researcher` to understand what unfamiliar attributes/error codes mean.
- **Strict output format** — data findings stated first; opinions only in a separate "My read" section with explicit confidence and basis.

## Other changes

- Register the skill in `.github/copilot-instructions.md` so the agent picks it up.
- Add a pointer from `kusto-analyst` SKILL.md so Kusto-query authors delegate the Aria-specific workflow back here.
- Fix two unrelated pre-existing typos in `kusto-analyst/SKILL.md` (corrupted MATS table row `+------=+|` and `isspecified` missing space).

## Tested via

Used the new workflow end-to-end on IcM 786276130 (`TD - ATS Devices W/ Success` Aria alert) earlier today. Multi-angle trend correctly attributed the metric drop to the 14.2.0 broker rollout instead of the alert grain itself, matching the historical pattern from sibling ICMs.

Fixes [AB#3589793](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3589793)


.github/skills/aria-alert-investigator/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+## When to use this skill vs. the others
+
+| Incident shape | Use |
+|----------------|-----|
+| Customer/partner reports auth failure, missing PRT, sign-out loop, etc. | `incident-investigator` |
+| You need to write Kusto queries for any reason | `kusto-analyst` (this skill calls into it) |
+| **IcM title is "Aria detected an incident in `<project>` for `<metric>`"** | **This skill** |
+| User says "investigate this Aria alert" / "health metric IcM fired" | **This skill** |
+
+Aria alerts are **statistical anomaly detections on telemetry**, not user-reported problems. The investigation pattern is fundamentally different — there is often no customer, no error chain, and no log file. The signal is "the curve moved" and the job is to determine whether the underlying data actually moved, and if so, why.
```
shahzaibj: They aren't anomaly detection actually - that's a different thing technically.

Our Aria alerts are simply comparing the value against a baseline and once it crosses it fires.

.github/skills/aria-alert-investigator/SKILL.md
```diff
@@ -0,0 +1,203 @@
...
+
+Do this whenever an unfamiliar attribute, error code, or operation appears in the data. The data only tells you *that* something moved; the code tells you *what it would mean* if it did.
+
+---
+
+## Workflow
+
+Execute steps in order. Do not skip steps.
+
+### Step 1 — Decode the metric → KQL slice (BLOCKING)
```
shahzaibj, line 82: What does BLOCKING mean here?

shahzaibj, line 82: What does BLOCKING mean here?

shahzaibj (approved):

.github/skills/aria-alert-investigator/SKILL.md
somalaya: it just means it is a necessary step.

somalaya: it just means it is a necessary step.

.github/skills/aria-alert-investigator/SKILL.md
somalaya: @copilot , can you take this suggestion and fix?

PR #436, opened 2026-05-20
Add Doze network-block repro test script, Fixes AB#3612535

rpdome:
[AB#3612535](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3612535)

Automated script to reproduce Android Doze network block on Broker auth from background callers. Requires SilentAuthReceiver from msal PR #2516. Enforces single-broker, checks battery exemption, reports results neutrally.

### DIFF
--- a/scripts/doze-repro-test.sh
+++ b/scripts/doze-repro-test.sh
@@ -51,7 +51,7 @@ echo "============================================="
 echo ""
 
 # --- Prerequisite checks ---
-echo "[1/8] Checking adb connection..."
+echo "[1/9] Checking adb connection..."
 if ! adb devices 2>/dev/null | grep -q "device$"; then
     echo -e "${RED}ERROR: No adb device connected.${NC}"
     echo "Connect a device via USB and authorize adb."
@@ -71,7 +71,7 @@ adb_cmd() {
 
 cleanup() {
     if [ "$DOZE_FORCED" = true ]; then
-        echo "[8/8] Cleaning up Doze..."
+        echo "[9/9] Cleaning up Doze..."
         adb_cmd shell dumpsys deviceidle unforce > /dev/null 2>&1 || true
         adb_cmd shell dumpsys battery reset > /dev/null 2>&1 || true
         DOZE_FORCED=false
@@ -81,7 +81,7 @@ cleanup() {
 
 trap cleanup EXIT INT TERM
 
-echo "[2/8] Checking SilentAuthReceiver is registered..."
+echo "[2/9] Checking SilentAuthReceiver is registered..."
 if ! adb_cmd shell "dumpsys package $RECEIVER_PKG" 2>/dev/null | grep -q "SilentAuthReceiver"; then
     echo -e "${RED}ERROR: SilentAuthReceiver not found in $RECEIVER_PKG.${NC}"
     echo "Install MsalTestApp."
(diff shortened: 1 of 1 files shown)


siddhijain (approved):

scripts/doze-repro-test.sh
```diff
@@ -0,0 +1,215 @@
...
+    exit 1
+fi
+if [ ${#INSTALLED_BROKERS[@]} -gt 1 ]; then
+    echo -e "${RED}ERROR: Multiple broker apps installed (${#INSTALLED_BROKERS[@]}):${NC}"
+    for i in "${!INSTALLED_BROKERS[@]}"; do
+        echo "  - ${INSTALLED_NAMES[$i]} (${INSTALLED_BROKERS[$i]})"
+    done
+    echo ""
+    echo "Uninstall all but one broker to get a deterministic test."
+    echo "With multiple brokers, MSAL's broker discovery picks one at runtime"
```
Prvnkmr337, line 125: **Device left in Doze if script fails mid-execution**

The script uses `set -euo pipefail`, so any failing command causes an immediate exit. The device state is modified at step [5/8]:

```bash
adb shell dumpsys battery unplug
adb shell dumpsys deviceidle force-idle
```

But the cleanup that reverses this only runs at step [8/8] — which is skipped if any command between steps 5–8 fails (e.g., `adb logcat -d`, `am broadcast`, or even `grep`/`sed` in the results section). This leaves the device stuck with battery unplugged + Doze forced, requiring manual recovery.

Consider adding a `trap` handler to guarantee cleanup:

```bash
DOZE_FORCED=false

cleanup() {
    if [ "$DOZE_FORCED" = true ]; then
        echo "Cleaning up Doze..."
        adb shell dumpsys deviceidle unforce > /dev/null 2>&1 || true
        adb shell dumpsys battery reset > /dev/null 2>&1 || true
    fi
}
trap cleanup EXIT
```

Then set `DOZE_FORCED=true` after successfully forcing Doze at step [5/8], and remove the manual cleanup at step [8/8] (the trap handles it).

scripts/doze-repro-test.sh
```diff
@@ -0,0 +1,215 @@
...
+RED='\033[0;31m'
+GREEN='\033[0;32m'
+YELLOW='\033[1;33m'
+NC='\033[0m'
+
+echo "============================================="
+echo " Doze Network Block Reproduction Test"
+echo "============================================="
+echo ""
+
```
Prvnkmr337, line 52: **SERIAL captured but never used**

The script captures the device serial at step [1/8]:

```bash
SERIAL=$(adb devices | grep "device$" | head -1 | awk '{print $1}')
```

But all subsequent `adb` commands don't use it. If multiple devices/emulators are connected, they'll fail with `adb: more than one device/emulator` — even though the script already knows which device to target.

Consider using `-s "$SERIAL"` throughout, or defining a helper:

```bash
ADB="adb -s $SERIAL"
$ADB shell dumpsys battery unplug
$ADB shell dumpsys deviceidle force-idle
# ... etc
```

This also makes the `trap cleanup` (from the other comment) work reliably when multiple devices are attached.

Prvnkmr337 (approved):
Few minor comment, but overall LGTM.

scripts/doze-repro-test.sh
rpdome: @copilot address this please

scripts/doze-repro-test.sh
rpdome: @copilot for simplicity, let's fail the test early if more than one device is connected.

PR #445, opened 2026-07-09
s360-reporter: auto-fetch prior report via WorkIQ Graph (AB#3683195), Fixes AB#3683195

shahzaibj:
Fixes [AB#3683195](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3683195).

## Why
Cesar''s star-5 skill-effectiveness survey for `s360-reporter` flagged the current Step 0b flow — where the user pastes last week''s report into a VS Code text box — as the top pain point. VS Code silently truncates large HTML pastes, which corrupts the previous-report map and causes items that existed last week to be mislabeled as *new* in this week''s diff.

## What
`SKILL.md`-only change. Flip Step 0b from paste-first to Graph-fetch-first.

- **New default (Method 1):** call the WorkIQ MCP''s Graph passthrough (`workiq-fetch`, runtime tool identifier `mcp_workiq_fetch`) with:
  ```
  GET /me/messages?$search="S360 Weekly Report"&$select=id,subject,body,sentDateTime,receivedDateTime&$top=5
  ```
  Filter/sort in code — `$search` can''t be combined with `$filter`. Drop messages < 3 days old (current cycle), take the most recent remaining, parse `body.content` with the existing Step 3a rules. Log a one-liner naming the picked email so the choice is auditable.
- **Fallback tier:** `workiq-fetch` -> `mcp_workiq_ask_work_iq` NL query -> `ask_user` paste (last resort). Paste stays available for the offline / Purview-blocked case but is never the default.
- **Step 3a** is reframed as parser-only, invoked from Step 0b — no fetch logic there anymore. All AB# / Bug / PBI / URL extraction rules preserved verbatim.
- **Prerequisites**: WorkIQ MCP promoted from *optional* to *required for default flow*.
- **Edge Cases**: added the three-tier fallback policy and the `$search`/`$filter` incompatibility note.
- **Sweep**: replaced remaining "if the user pasted" references in Step 1e / resolved section with "from the fetched prior report".

No JS/parser code changes — parsing was prose-only and the same rules now serve all three sources.

## Verification
Skill-doc change; no build/test. Suggested next-OCE dry run:
1. Run the skill; confirm no paste prompt appears at the top of the flow.
2. Confirm the log line prints something like `Fetched prior report: "S360 Weekly Report — <date>" (sent ...Z) — parsed N items.`
3. Confirm the diff in Step 5 doesn''t mark previously-tracked items as *new*.
4. To exercise the fallback, temporarily disable the WorkIQ MCP in `.vscode/mcp.json` and re-run — the skill should degrade to `ask_work_iq`, and finally to the paste prompt.

## Open item
The user request names the tool `workiq-fetch`. The `@microsoft/workiq` MCP is installed but not connected in this session, so I couldn''t live-probe the exact runtime identifier. SKILL.md documents both the raw name (`workiq-fetch`) and the likely runtime form (`mcp_workiq_fetch`) — please reconcile on first live run if the identifier differs.

Co-authored-by: Copilot App <223556219+Copilot@users.noreply.github.com>

### DIFF
--- a/.github/skills/s360-reporter/SKILL.md
+++ b/.github/skills/s360-reporter/SKILL.md
@@ -16,7 +16,10 @@ HTML email report.
 - **S360 MCP Server** must be running (configured in `.vscode/mcp.json` as `s360-breeze-mcp`)
 - **ADO MCP Server** must be running (for PBI creation and lookup)
 - **M365 User MCP** (`m365-user`) — for dynamic team member discovery via org chart
-- **WorkIQ MCP Server** (optional) — used as fallback for pulling last week's email if the user doesn't provide it
+- **WorkIQ MCP Server** (`workiq`) — **required for the default flow**. Used in Step 0b to
+  auto-fetch last week's S360 report from the user's mailbox via a Microsoft Graph
+  passthrough (`workiq-fetch` on `/me/messages`). If WorkIQ is unavailable, the skill
+  falls back to a manual paste — see Step 0b for the full fallback tier
 - Read the **Outlook HTML report prompt** at `{{VSCODE_USER_PROMPTS_FOLDER}}/outlook-html-report.prompt.md`
   for HTML rendering rules before generating the report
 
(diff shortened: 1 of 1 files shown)


shahzaibj:
Not planning to merge at the moment

PR #449, opened 2026-07-18
Enhance Copilot Review Analyst skill: v3 newsletter report + adoption-signal & fairness methodology, Fixes AB#3696013

shahzaibj:
## Summary

Bakes this cycle's learnings into the **copilot-review-analyst** skill (`.github/skills/copilot-review-analyst`). Two themes: a redesigned **v3 newsletter** report + hardened Outlook rendering, and a stronger, fairer **classification methodology**.

Fixes [AB#3696013](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3696013)

## Report / rendering
- **v3 newsletter Outlook report** (`assets/…-Outlook.html`): 1200px hybrid-fluid shell, edge-to-edge newsletter masthead (inset hero + 1120px VML width), big-number KPI hero grid, single-row meta-chips, and a Methodology section. Committed as the **anonymized** sample, single-quoted (escaping-free).
- **Headline leads with adoption %** (Helpful & adopted), not precision — precision is near-100% by construction once Declined/Unresolved are excluded, so leading with it overstates the story.
- **Outlook rendering learnings** documented in `references/report-formatting.md` § *v3 Newsletter Rendering*:
  - Meta-chips must be **one single-row table** — separate `inline-block` tables stack vertically in classic Outlook.
  - Body prose needs `line-height` **plus** `mso-line-height-rule:exactly` or the Word engine collapses leading.
  - **Edge-to-edge frame**: whiten the outer canvas `bgcolor` **and** inline CSS (Word engine reads `bgcolor`).
  - **Newsletter masthead**: inset hero to the text column + VML fallback width = container − 2×gutter.
  - **Delivery**: Graph draft preferred, with an Outlook COM fallback when WorkIQ/Graph is unavailable.

## Classification / methodology
- **Declined verdict + Copilot precision** = `Helpful / (Helpful + Incorrect)`; Declined is neutral and never counted against Copilot.
- **Force-push confound validation** before scoring precision (a correct comment on a rewritten snapshot no longer looks like a false positive).
- **Engineer-intention nuance**: correct-but-declined feedback is a judgment call, not a Copilot error.
- **New adoption signals**: reaction + apply-suggestion detection (silent adoption), merged-only PR filter, Phase 0 preflight coverage audit, review-coverage split, self-consistency gate, and a PR triage table.

## Notes
- The standard `.md`/`.html` assets remain structural references (illustrative prior-run data — computed fresh each run).
- No product code touched; skill/docs only.


fadidurah:
PR Body mentioned a .gitignore addition, but i'm not seeing it in the files, is it stale?

fadidurah (approved):
LGTM

.github/skills/copilot-review-analyst/scripts/append-history.ps1
```diff
@@ -166,8 +175,13 @@ $snapshot = [ordered]@{
     responseRate       = $responseRate
     helpful            = [ordered]@{ count = $confirmedHelpful; pct = $helpfulPct }
     notHelpful         = [ordered]@{ count = $confirmedNotHelpful; pct = $notHelpfulPct }
+    notHelpfulIncorrect = [ordered]@{ count = $confirmedNotHelpfulIncorrect; pct = $notHelpfulIncorrectPct }
+    declined           = [ordered]@{ count = $declinedCount; pct = $declinedPct }
     unresolved         = [ordered]@{ count = $unresolved; pct = $unresolvedPct }
```
somalaya, line 176: unknown verdict is dropped from the snapshot -  buckets won't reconcile to total.

$unknownCount is computed but never added to $snapshot. The emitted buckets are helpful + notHelpfulIncorrect + declined + unresolved, so any unknown comment makes them sum to less than total.

unknown does occur: in final-classification.ps1, a replied comment absent from ReplyVerdictsFile/OverridesFile stays "unknown", and that script's self-consistency gate treats unknown as a canonical bucket — so omitting it here makes the history inconsistent with the classification output.

.github/skills/copilot-review-analyst/scripts/append-history.ps1
```diff
@@ -54,102 +56,109 @@ if (Test-Path $precisePath) {
     $precise = Get-Content $precisePath -Raw | ConvertFrom-Json
```
somalaya: $precise / precise.json is now dead code.

This PR removed the $preciseMap re-derivation , but precise.json is still loaded into $precise here and never referenced again anywhere in the file. The .SYNOPSIS and .PARAMETER InputDir also still say the script "Reads final_classification.json and precise.json."

Suggest dropping the load and the $precisePath line, and updating the doc comments, so the script no longer implies a dependency it doesn't have.

.github/skills/copilot-review-analyst/scripts/append-history.ps1
shahzaibj: Good catch — fixed in 2a03435. `$unknownCount` was computed but never written to the snapshot, so any comment that stayed `unknown` made the additive buckets under-sum vs `total` and left history.json inconsistent with `final-classification.ps1`'s canonical taxonomy.

The snapshot now emits an `unknown = { count, pct }` bucket (after `unresolved`), prints an Unknown line in the console summary, and I added a self-consistency assertion that warns if `helpful + incorrect + declined + unresolved + unknown` doesn't reconcile to `total`. (`notHelpful` stays a derived combined field, intentionally excluded from that sum.) Verified with a synthetic run containing an `unknown` — buckets reconcile 5/5, no warning.

shahzaibj:
@fadidurah you're right, it was stale. The `.gitignore` change was unrelated to this PR (it snuck in from an auto-merge), so I dropped it in 1e6dac5 — the diff now touches only the skill files. I've also updated the PR description to remove that line. Thanks for flagging.

somalaya (approved):

.github/skills/copilot-review-analyst/scripts/append-history.ps1
shahzaibj: You're right — dead code. Removed in 60f214c: dropped both `$precisePath` and the `$precise` load block, and updated `.DESCRIPTION` + `.PARAMETER InputDir` to reference only `final_classification.json`.

I kept the one rationale comment noting we deliberately *don't* re-derive silent-helpful from precise.json — that documents the absence of the dependency rather than implying one, so it shouldn't be misleading. Verified the script runs end-to-end with no precise.json present and buckets still reconcile to total.

PR #452, opened 2026-07-24
Add android-e2e-tester skill (consolidated reports, Attachments, explicit-only trigger), Fixes AB#3696527

wzhipan:
## Summary

Adds the **`android-e2e-tester`** VS Code skill — an end-to-end test-automation workflow for Android Auth features (MSAL, Broker, Common, ADAL, Authenticator) on emulators and real devices, driven from Azure DevOps test cases. This branch also contains the latest refinements to the skill's reporting and triggering.

The PR is scoped entirely to `.github/skills/android-e2e-tester/` (20 files, additions only against the merge base).

**Work item:** [AB#3696527](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3696527)

## Use cases

- **End-to-end feature development.** After an initial implementation lands, the agent validates the feature with a real E2E run, **finds bugs and auto-iterates on fixes**, or escalates a hard gap that blocks completion — so a feature can be driven to "works on device" without a human babysitting each step. Greatly increases Copilot-agent autonomy and cuts engineer effort and context-switching.
- **Autonomous ADO test-case execution (self-serve bug bash).** Runs manual Broker/MSAL/Authenticator test cases without an engineer doing them by hand — it **pulls the test case from ADO**, **gets accounts and passwords from the LAB API**, runs the scenario, and **generates a test report** afterward. It also **suggests improvements to the test steps** so the ADO cases get clearer over time.
- **Additional use cases the same engine enables:**
  - **Parallel, multi-device execution** — splits a suite across attached emulators/devices (one sub-agent per case) for a scaled regression / bug-bash pass, then rolls up a suite-level summary.
  - **Multi-config validation** — runs every test point per case, mapping `LocalFlights` → Local and plain configs → ECS builds, so a feature is checked against both flighting states in one go.
  - **Cross-device compatibility** — the same scenario on an emulator (where it can inject fingerprint / advance the clock) and on real hardware.
  - **Audit-ready reporting** — self-contained HTML + Markdown reports with steps, evidence, and screenshots for sign-off and compliance.
  - **Institutional-knowledge capture** — every run enriches a shared references library (common blockers, LAB API usage, device/PIN handling), so the next run is faster and the whole team benefits.

## What the skill does
- Builds a device pool (emulators + adb-connected hardware), leases devices so parallel runs don't collide, builds/installs the right test APKs, drives the UI, and verifies via logcat.
- Delegates each run to a sub-agent and supervises it; when several test cases and devices are available, splits cases across devices and runs them in parallel (one sub-agent per case).
- For ADO test cases, always writes an **HTML + Markdown test report**; a batch also gets an overall run summary.

## Recent refinements included here
- **Consolidated multi-test-point reports** — one report per test case with a section per test point (`testPoints[]`), a derived overall verdict, and a suite summary that expands points into one row each. Backward compatible with the flat single-point run.json.
- **Proposed-steps section** with an **Attachments** column, authored once at case level and generic across all test points.
- **LAB account freshness gate** — poll a new temp user ≤ 3 min; if not sign-in-able, recreate or reuse a < 30-min-old user (never a password reset).
- **Trigger restricted to explicit user requests** — the skill now starts **only when the user explicitly asks to run an E2E test** (e.g. "run the E2E test", "run this ADO test case"). It no longer auto-starts after a feature finishes development, when a coding-agent PR lands, or as an implicit step of another workflow.

## Sample test report

A consolidated report from a sample run of the "Cross-Apps SSO with System Browser" ADO case (**TC 497038**) across **both** test points (ECS + Local/LocalFlights builds) → **PASS**, with per-test-point sections, evidence, and screenshots:

- [TestReport.html](https://onedrive.cloud.microsoft/:u:/a@2th7ra99/r/_layouts/15/onedrive.aspx?id=%2Fa%402th7ra99%2FDocuments%2FAttachments%2FAndroid%2De2e%2Dtester%2Dskill%2FSample%2Dtest%2Dreport%2Ftc497038%2FTestReport%2Ehtml&parent=%2Fa%402th7ra99%2FDocuments%2FAttachments%2FAndroid%2De2e%2Dtester%2Dskill%2FSample%2Dtest%2Dreport%2Ftc497038&share=cQrel8LVBxoUQJ2efL8H%2DQweEgUCV6xIDl0UPQVSXyGVa%5FhLOw)

## Notes
- All scripts parse under Windows PowerShell 5.1 and pwsh 7.x; reference-doc links/anchors resolve.
- No product/library code is touched — this is skill tooling and documentation only.

Co-authored-by: Copilot <223556219+Copilot@users.noreply.github.com>


.github/skills/android-e2e-tester/SKILL.md
```diff
@@ -0,0 +1,571 @@
...
...e feature on a device or emulator', 'does this work on a device', 'run this ADO test case', or 'try the sign-in flow'. Do NOT auto-start it after a feature finishes development, when a coding-agent PR lands, or as an implicit step of any other workflow; a human has to ask for an E2E run. Discovers what to test from user-provided steps, an ADO Test Case work item, a known test-steps file, the session's design spec, or the implementation diff — feature-specific steps live in the test case, not the skill. Auto-performs inputs the AI can handle (typing lab test credentials, tapping buttons, granting permissions, simulating a fingerprint, provisioning lab accounts via the LAB API), mocks unavailable dependencies and sets feature flags via temporary code changes when needed, and asks the user when the intent is unclear or a step is a real blocker (push MFA, hardware, missing credentials, implementation gaps). Checks logs to decide pass/fail, drives a fix-and-retest loop until it passes, and — for ADO test cases — always writes an HTML + Markdown test report at the end. Prefers an emulator when a step needs an injectable fingerprint/biometric (App Lock, number-match with biometric gate)."
```
cacosta33: 🟡 **Suggestion:** this `description` is 2,098 characters — about 4.5× every other skill in the repo (next largest is `copilot-review-analyst` at 607; the median across the other 17 is ~470).

Because `description` is loaded for every skill-selection decision in every session, an outlier this large is a permanent context cost paid by all sessions, not just E2E ones. It can also work against itself: the explicit-only guard ("do NOT auto-start after a feature finishes development…") is the highest-value sentence here, and it's currently surrounded by a lot of prose that reads more like body content.

Suggest trimming to ~2–4 sentences — what the skill does, plus the explicit-only guard — and moving the rest into the body under `## When this runs`, which already covers this ground.

.github/skills/android-e2e-tester/scripts/labapi.ps1
```diff
@@ -0,0 +1,326 @@
...
+        "--remote-debugging-port=$port", "--user-data-dir=$prof",
+        '--no-first-run', '--no-default-browser-check', $Endpoint)
+    try {
+        $deadline = (Get-Date).AddSeconds([Math]::Max(12, $TimeoutSec))
+        $wsUrl = $null
+        while ((Get-Date) -lt $deadline -and -not $wsUrl) {
+            Start-Sleep -Milliseconds 400
+            try { $targets = Invoke-RestMethod "http://127.0.0.1:$port/json/list" -TimeoutSec 3 } catch { continue }
+            # Only the real content page (http/https); skip edge:// dialogs (e.g. sync-confirmation).
+            $page = $targets | Where-Object { $_.type -eq 'page' -and $_.url -like 'http*' } | Select-Object -First 1
```
cacosta33: 🟡 **Suggestion:** the headless-Edge/CDP path isn't safe under concurrency, which sits awkwardly with parallel multi-device execution being a headline feature of the skill.

Three things compound when two `labapi` calls overlap:

- **Shared profile dir by default** — L144–145 uses a fixed `$env:TEMP\labapi_edge_cdp_profile` unless `-Fresh`. Two concurrent Edge launches against the same `--user-data-dir` generally don't both start; the second hands off to the first and exits.
- **Random port, no bind check** — L146 draws from a 299-port range, and nothing verifies Edge actually bound *that* port.
- **No target validation** — L158 takes the first `type -eq 'page'` with an `http*` URL and never checks it corresponds to `$Endpoint`. If it attaches to another instance, the call silently returns that instance's page content as this call's API response.

`Stop-Process` in the `finally` (L172) can then also tear down a browser the other run is mid-eval on.

The silent-wrong-answer case is the dangerous one. Two small changes remove it: always use a unique profile dir (drop the shared-dir branch, or default `-Fresh` on), and assert `$page.url` starts with `$Endpoint` before taking `webSocketDebuggerUrl`.

.github/skills/android-e2e-tester/scripts/labapi.ps1
```diff
@@ -0,0 +1,326 @@
...
+
+function Show-Result {
+    param($Result)
+    if ($Result.Json) {
+        $Result.Json | ConvertTo-Json -Depth 8
+    }
+    else {
+        Write-Host $Result.Raw
+    }
+}
```
cacosta33, line 292: 🟡 **Suggestion:** `Show-Result` does `$Result.Json | ConvertTo-Json -Depth 8` and prints the whole response, and it's used by `create-user`, `reset`, `enable-policy`, `disable-policy`, and `delete-device`.

To be clear, **today's responses are safe** — `references/lab-api.md:142–152` shows create-user returns `passwordUri` / `credentialVaultKeyName`, i.e. URIs and key *names*, not values. The concern is structural: the skill's stated invariant is that passwords never reach the chat transcript, and right now that invariant depends on an external service's response shape never gaining a `password`/`secret` field.

A ~3-line key filter (`password|secret|token|key`) before the dump would make the guarantee hold by construction rather than by coincidence.

.github/skills/android-e2e-tester/scripts/labapi.ps1
```diff
@@ -0,0 +1,326 @@
...
+function Save-DpapiSecret {
+    # Cache a plaintext into the local DPAPI store in the exact format scripts/secrets.ps1 uses, so it
+    # resolves via `-SecretRef <Name>`. Nothing is printed here; the caller emits only a masked summary.
+    param([Parameter(Mandatory)][string]$Name, [Parameter(Mandatory)][string]$Value)
+    $dir = Join-Path $env:USERPROFILE '.android-e2e-secrets'
+    if (-not (Test-Path $dir)) { New-Item -ItemType Directory -Force -Path $dir | Out-Null }
+    $ss = ConvertTo-SecureString $Value -AsPlainText -Force
+    $enc = ConvertFrom-SecureString $ss    # DPAPI encrypt (per-user, per-machine)
+    Set-Content -Path (Join-Path $dir ("{0}.sec" -f $Name)) -Value $enc -NoNewline -Encoding ASCII
+}
```
cacosta33, line 336: 🟢 **Nit:** `Save-DpapiSecret` re-implements the `secrets.ps1` store rather than reusing it — it hardcodes `'.android-e2e-secrets'` a second time (the first is `secrets.ps1:80 Get-StoreDir`) and re-derives the `.sec` / `ConvertFrom-SecureString` format.

The comment says *"in the exact format scripts/secrets.ps1 uses"*, which names the coupling without enforcing it. Worth noting `secrets.ps1:196` already has a dot-source guard (`if ($MyInvocation.InvocationName -ne '.')`) specifically so its functions can be reused, so dot-sourcing looks like the intended pattern here.

Low impact today; the risk is silent drift if the store path or encoding ever changes on only one side.

PR #455, opened 2026-08-02
Add Authenticator telemetry report to OCE skill; align noise gate to the rolling window

shahzaibj:
## What

Adds the **Authenticator app** telemetry report to the existing `oncall-weekly-telemetry-report` skill, and fixes a noise-gate bug that was hiding real regressions from on-call engineers.

One slash-command invocation now produces **two separate reports** plus a combined index. Modes: `both` (default) | `broker` | `authapp`.

| Mode | Output |
|---|---|
| `broker` | `oncall-wow-report-<curEnd>.html` |
| `authapp` | `authapp-wow-report-<curEnd>.html` |
| `both` (default) | both, **plus** `oce-index-<curEnd>.html` |

The skill was restructured into a **router** (`SKILL.md`) plus two app playbooks, which run as parallel sub-agents in `both` mode. The two clusters' Kusto conventions are incompatible (Broker requires HLL device counting; Authenticator's `sum(*DCount)` is correct there and the Broker rule is actively wrong), so keeping them in separate contexts is deliberate — not just a size optimisation.

## Why — the noise-gate bug

The novelty classifier graded **Sunday-aligned calendar weeks** cut off at `startofweek(curEnd)`, while the report displays a **rolling 7-day window**. On the 2026-08-01 run the gate's "current" week was 07/19–07/26 against a report window of 07/25–08/01 — **one day of overlap** — so anything that turned in the final ~6 days was structurally invisible to the gate.

Two real risers were suppressed as *"ONGOING — do not re-triage"*:

| code | report said | classifier saw |
|---|---|---|
| `authorization_pending` | **+63.2%** | −37.1% |
| `expired_token` | **+26.7%** | −51.0% |

Both were codes the on-call engineer had already spotted by hand — which is how the bug was found.

**Fix:** bucket with `bin_at(t, 7d, datetime(curEnd))` everywhere. The newest bucket becomes exactly `[curEnd-7d, curEnd)` — the report's own window — so **classifier WoW equals displayed WoW by construction**. Every bucket is a complete 7 days; the 4-day stub moves to the *oldest* bucket where `--start` drops it, leaving 8 clean rolling weeks. `TREND_CLASS_END` and `--include-partial-end` are obsolete (the latter kept as a no-op).

This is **not** "just more alerts" — alignment removes phantom signal too. `access_denied` was ACCELERATING under calendar weeks but is actually **−53.2%**, and got correctly demoted. The attention set went 4 → **5** keys, not 4 → 15.

## Why — the second, independent defect

A key whose 60-day peak sits under the peak floor (Broker 10,000 / AuthApp 1,000 devices) is **excluded from classification outright**, so it can carry a red pill forever while the attention section says "quiet week".

`Passkey WebAuthN Registration` shipped `tag-bad` (−1.27 pts, worst delta in the table) directly above the words *"Quiet week — 0 NEW or ACCELERATING"*. Both statements were true; it peaks at **732** devices. **No bucketing change fixes this.**

Added a **reconciliation rule**: every `tag-bad`/`tag-warn` row must be either promoted into attention **or** named in a muted `.reconcile-note` giving the reason — tested in order: (1) below the classification floor, (2) within its own normal band, (3) ONGOING and flat. Enforced by new **`validate-report.ps1` check 19**, which hard-fails an unreconciled pill.

Also **reordered the Authenticator report** so *Needs attention* sits above the scenario scoreboard, matching the Broker report and the order an OCE actually reads in.

## Verification

Regenerated end-to-end against **live Kusto** for `curEnd = 2026-08-01`, then independently verified the HTML rather than trusting the generating agent's self-report:

**Broker** (108,459 bytes)
- `authorization_pending` — ACCELERATING in attention, +63.3% (172 K → 281 K devices) ✅
- `expired_token` — ACCELERATING in attention, +26.9% (86.3 K → 109 K devices) ✅
- `access_denied` — correctly **absent** ✅
- 3 visible attention rows — signal added without noise ✅
- `429` / `temporarily_unavailable` explained in a reconcile note ✅

**Authenticator** (60,688 bytes)
- Section order confirmed: health → **Needs attention** → **Scenario scoreboard** → 60-day trend ✅
- Genuine quiet week preserved (`NEW=0 ACCELERATING=0`) — no manufactured findings ✅
- Exactly 2 red/amber pills, both named in the single reconcile note — perfect 1:1 ✅

**Both:** `All hard checks passed`, 0 `U+FFFD`. The protected `2026-07-31` before-fixture still fires exactly its expected 3 hard FAILs, so the validator hasn't been loosened.

## Notes for reviewers

- **Skill-only change.** All 29 files are under `.github/skills/oncall-weekly-telemetry-report/`. No product/library code touched.
- Reports are written to `$env:USERPROFILE\android-oce-reports\` — deliberately **outside** the workspace so they can never be committed. None are in this PR.
- The branch is 2 commits behind `master`; the merge-base diff is clean, no rebase needed.


shahzaibj:
No longer needed

